PR #85, opened 2017-03-07
ViewModelLocator now connects view models to views, rather than NavigationService.

davidbritch:
Uses an auto-wiring convention-based approach.


CESARDELATORRE:
Thanks for this! - We'll follow up and review this pull request.
Btw, **congrats for being the first pull-request from the community to this project**!! :)
Want to send me you physical address to cesardl at microsoft.com so I send you a .NET T-Shirt from the team? :)

PR #106, opened 2017-03-09
Removes Docker profile from launchSettings.json

dazhao-msft:
The Docker profiles are out-of-date and no longer used by Docker tools. The change is to remove them to avoid issues with VS 2017 RTW.


CESARDELATORRE:
Thanks David! :)

PR #108, opened 2017-03-09
Revert "Removes Docker profile from launchSettings.json"

CESARDELATORRE:
Reverts dotnet/eShopOnContainers#106


CESARDELATORRE:
Testing, looks like there are errors, might or might be not due to this original pull request

PR #110, opened 2017-03-09
Update Compiler package

BillWagner:
This is necessary for C# 7 feature support.
See the pattern match is statement in AutoFacHttpModule.

@CESARDELATORRE this same change (update Microsoft.Net.Compilers package to 2.0 would likely be necessary for the MVC, SPA, and WebAPI packages to support the new language features.  I can do that on the dev branch if you'd like.


CESARDELATORRE:
Cool! We need to add C# 7 features to the projects! :)

Issue #66, opened 2017-02-26
Enforce Repository per AggregateRoot


ardalis:
Currently nothing in the code uses `IAggregateRoot`. One way to have the code better enforce the DDD convention that only Aggregate Roots should have Repositories would be to implement a generic repository type. Since our repositories aren't implementing any standard set of CRUD methods (currently it just exposes UnitOfWork) the change could simply be like this, in IAggregateRepository:

```
    public interface IAggregateRepository<T> where T:IAggregateRoot
    {
        IUnitOfWork UnitOfWork { get; }
    }
```

It's worth considering that `IAggregateRepository` is redundant as a name at that point. Since the rule is that only Aggregates (their roots) have repositories, the interface can simply be `IRepository`.

FDUdannychen:
+1 I'm going to ask the same question, I think `IOrderRepository : IRepository<Order>` makes more sense than `IOrderRepository<T> : IRepository<T>` since we don't need `IOrderRepository<Buyer>`.

CESARDELATORRE:
Agree. This is pending. Need to refactor it.

CESARDELATORRE:
@FDUdannychen  Btw, pull requests are accepted and appreciated. This could be an easy one.
We're heads down in the dev-branch focusing on Domain-Events, Integration Events, resiliency (retries/circuit breaker, etc.) and continuing with the Guide/eBook. 
But we'll try to refactor this issue when possible if we don't get any pull request. :)
And thanks for your feedback confirming this issue! :)

dsrodenas:
Pull request #121 created to change the signature to `IOrderRepository : IRepository<Order>`. This way we have `public class OrderRepository :  IOrderRepository `.

CESARDELATORRE:
Closing this issue.

PR #120, opened 2017-03-20
use non generic repository interface

FDUdannychen:
`IOrderRepository` makes more sense than `IOrderRepository<Order>` since we don't need `IOrderRepository<Buyer>`.


eiximenis:
LGTM!

eiximenis:
Merged.
This fixes #66

Issue #76, opened 2017-03-04
No retry logic at Catch Exceptions but when starting the services


CESARDELATORRE:
At the CatalogContextSeed.cs at the Catalog.API microservice.
Related to https://github.com/dotnet/eShopOnContainers/issues/75 
This retry in the Catch should be removed:
```
            catch (Exception ex)
            {
                if (retryForAvaiability < 10)
                {
                    retryForAvaiability++;
                    var log = loggerFactory.CreateLogger("catalog seed");
                    log.LogError(ex.Message);
                    await SeedAsync(applicationBuilder, loggerFactory, retryForAvaiability);
                }
            } 
```
Better have a retry loop at the **Startup.cs** before starting anything at the service so it waits until the db_container is ready. It doesn't need to be here in a catch().
----> Also, we will change to the HealthCheck lib when available from the team and will use it at the **Startup.cs** , too.

CESARDELATORRE:
Solved. Closing.

Issue #75, opened 2017-03-03
Pokemon exceptions


BillWagner:
Before final release, we should remove many of the catch {} code in the reference application.

That "catch all exceptions" without a real strategy to fix them is an antipattern. it's also one that generates very strong feelings:  https://github.com/dotnet/csharplang/issues/220

PR #122, opened 2017-03-20
Removed the catch with retry logic from CatalogContextSeed 

ramonTC:
Removed the catch with retry logic from CatalogContextSeed and create a retry loop at startup for db availability #76 #75


CESARDELATORRE:
It works. I just had an issue with "Docker for Windows" itself. Re-starting Docker for Windows solved it and this pull request is working.

ramonTC:
Perfect!

PR #123, opened 2017-03-20
Fix #61 Async suffix for namig convention.

### LINKED ISSUE #61: Async Suffix Naming Convention
See: https://msdn.microsoft.com/en-us/library/hh873175.aspx

Some interfaces and implementations should probably be renamed to include the `Async` suffix. For example:
https://github.com/dotnet/eShopOnContainers/blob/master/src/Services/Basket/Basket.API/Model/IBasketRepository.cs#L10-L12

dsrodenas:
Fix for #61 As requested in the issue simple change of names to fit the naming convention for TAP. 

Ordering repositories will be changed to async when we migrate to ef core 1.1.1


CESARDELATORRE:
Tested this PR. ok.

Issue #72, opened 2017-03-02
SPA and latest NPM version issue? (Confirm)


CESARDELATORRE:
[This issue does not have the highest priority as with the LTS version of NODE there's no issue]
I might be wrong in these tests, but just in case..
Usually we're working with the 6.9.3 LTS version of NODE (currently 3.10.10 of NPM) to build the SPA app with: npm run build:prod

However, by mistake I installed the latest version of NPM (by installing the v7.4.0 of Node) and seems that with that version npm run build:prod is not able to work properly.
I tried running first the following:
npm install
npm rebuild node-sass

But no success. Please, when possible test the SPA app with the latest version of NODE and NPM so we know what's the current state and compatilbility.

eiximenis:
I tested with node 7x several times with no issues.
So, i'll close the issue.

PR #124, opened 2017-03-20
Resolved SPA incompatibilities with dependencies 

ramonTC:
#72  Resolved SPA incompatibilities with dependencies when installing a newer node version.
The actual version of node-sass dependency was not compatible with newest node versions so that the node-sass lib version used has been updated


CESARDELATORRE:
PR Tested. ok.

PR #125, opened 2017-03-20
Xamarin mobile client updates.

davidbritch:


CESARDELATORRE:
David, on question. Are you testing this PR code with the Xamarin app consuming the Docker containers as explained here:
https://github.com/dotnet/eShopOnContainers/wiki/03.-Setting-the-eShopOnContainers-solution-up-in-a-Windows-CLI-environment-(dotnet-CLI,-Docker-CLI-and-VS-Code) 

Basically, when trying the Xamarin app to consume the containers you need to configure the real external IP of your Docker Host here:
https://github.com/dotnet/eShopOnContainers/blob/master/.env 

And deploy the containers using the docker-compose.prod.yml:
https://github.com/dotnet/eShopOnContainers/blob/master/docker-compose.prod.yml
Which is using that external IP that can be consumed from the Xamarin app in the same WiFi, for instance.

We had a few issues the other day when trying your original PR and I'm not sure if there was any incompatibility with the new code and when consuming for real the microservices deployed as Docker containers  instead of just testing the Xamarin app with its "mock data feature". 

Please, confirm.

PR #126, opened 2017-03-20
C# 7 language feature updates

BillWagner:
@CESARDELATORRE Let me know if this should be targetting master instead of dev. I've seen  ongoing work in each branch.

*Controllers/AccountController:*
Replace single line methods with expression bodied members.
Includes constructor, C# 7 feature.

*Controllers/CatalogController:*
Remove unnecessary ToString() call.
Replace single line methods with expression bodied members.

*Extensions/HttpClientExtensions:*
Replace single line methods with expression bodied members.

*Extensions/SessionExtensions:*
Replace single line methods with expression bodied members.

*Services/BasketService:*
Remove unnecessary ToString() calls.
Add ?? to simplify conditional initialization
Use TryGetValue and out variable initialization to simplify Quantity
calculation

*Services/CatalogService:*
Use Value<T> generic method instead of dynamic types.
There is a lot of overhead for dynamic and it seems overkill here.

*Services/IdentityParser:*
Use the pattern matching is expression.
Refactor the LINQ queries to enumerate the collection (and create an
enumerator)
only once. The previous code had 3 enumerations.

*Services/Utilities/HttpApiClient:*
Remove the 'async' modifier and 'await' statements from methods where
the only asynchronous statement is the final statement of the method,
and
the task from the underlying method can be returned.

*Services/Utilities/HttpApiClientWrapper:*
Use expression bodied members where applicable.
Remove the 'async' modifier and 'await' statements from methods where
the only asynchronous statement is the final statement of the method,
and
the task from the underlying method can be returned.

*ViewComponents/Cart:*
Use expression bodied members where applicable.

*ViewComponents/CartList:*
Use expression bodied members where applicable.
Remove the 'async' modifier and 'await' statements from methods where
the only asynchronous statement is the final statement of the method,
and
the task from the underlying method can be returned.

*ViewModels/Annotations/CardExpiration:*
Use the out variable initializer to simplify the validation of the
card expiration date.

*ViewModels/Basket:*
Use property initializer syntax instead of constructor body

*ViewModels/CatalogViewModels/IndexViewModel:*
Use expression bodied property to return the calculated 'Disabled'
property

*ViewModels/Order:*
Use property initializer syntax.

### DIFF
--- a/src/Web/WebMVC/Controllers/AccountController.cs
+++ b/src/Web/WebMVC/Controllers/AccountController.cs
@@ -11,16 +11,11 @@ namespace Microsoft.eShopOnContainers.WebMVC.Controllers
     public class AccountController : Controller
     {
         private readonly IIdentityParser<ApplicationUser> _identityParser;
-        public AccountController(IIdentityParser<ApplicationUser> identityParser)
-        {
+        public AccountController(IIdentityParser<ApplicationUser> identityParser) => 
             _identityParser = identityParser;
-        }
 
-        public ActionResult Index()
-        {
-            return View();
-        }
-
+        public ActionResult Index() => View();
+        
         [Authorize]
         public IActionResult SignIn(string returnUrl)
         {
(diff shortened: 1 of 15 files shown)


CESARDELATORRE:
Let's keep it in DEV and it'll be merged with the new features, eventually. 
In MASTER we're mostly fixing bugs.
I'll test this PR before merging.

eiximenis:
Merged!

PR #128, opened 2017-03-21
Refactor and update to C# 7 constructs

BillWagner:
Controllers/BasketController:
Use Expression Bodied Members.
Delete() should not be 'fire and forget'

Controllers/HomeController:
Expression bodied constructor

IntegrationEvents/EventHandling/ProductPriceChangedIntegratedEventHandler:
Use valueTask to minimize allocations for 'fast path' tasks. One might
believe that
most users would not have the changing item in the basket at any given
time.
Using ValueTask instead of Task, this minimizes allocations to only when
async
work must actually take place.

IntegrationEvents/EventHandling/ProductPriceChangedIntegratedEvent:
Make the properties ReadOnly

Model/Basket:
Use property initializer

Model/RedisBasketRepository:
Remove redundant ToString()
Use elvis operator instead of statement null check

Startup:
remove redundant ToString()

### DIFF
--- a/src/Services/Basket/Basket.API/IntegrationEvents/EventHandling/ProductPriceChangedIntegrationEventHandler.cs
+++ b/src/Services/Basket/Basket.API/IntegrationEvents/EventHandling/ProductPriceChangedIntegrationEventHandler.cs
@@ -33,7 +33,7 @@ namespace Microsoft.eShopOnContainers.Services.Basket.API.IntegrationEvents.Even
         {
             var itemsToUpdate = basket?.Items?.Where(x => int.Parse(x.ProductId) == productId).ToList();
 
-            if ((itemsToUpdate != null) && (itemsToUpdate.Any())
+            if ((itemsToUpdate != null) && (itemsToUpdate.Any()))
             {
                 foreach (var item in itemsToUpdate)
                 {
--- a/src/Services/Basket/Basket.API/IntegrationEvents/EventHandling/ProductPriceChangedIntegrationEventHandler.cs
+++ b/src/Services/Basket/Basket.API/IntegrationEvents/EventHandling/ProductPriceChangedIntegrationEventHandler.cs
@@ -5,6 +5,7 @@ using System;
 using System.Linq;
 using System.Threading.Tasks;
 
+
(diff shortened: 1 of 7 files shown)


src/Services/Basket/Basket.API/IntegrationEvents/EventHandling/ProductPriceChangedIntegrationEventHandler.cs
```diff
@@ -5,6 +5,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 
+
```
jzeferino: Empty line should be removed.

src/Services/Basket/Basket.API/IntegrationEvents/EventHandling/ProductPriceChangedIntegrationEventHandler.cs
```diff
@@ -16,7 +17,7 @@ public ProductPriceChangedIntegrationEventHandler(IBasketRepository repository)
             _repository = repository ?? throw new ArgumentNullException(nameof(repository));
         }
 
-        public async Task Handle(ProductPriceChangedIntegrationEvent @event)
+         public async Task Handle(ProductPriceChangedIntegrationEvent @event)
```
jzeferino: Space before method should be removed.

src/Services/Basket/Basket.API/IntegrationEvents/EventHandling/ProductPriceChangedIntegrationEventHandler.cs
```diff
@@ -32,7 +33,7 @@ public ProductPriceChangedIntegrationEventHandler(IBasketRepository repository)
         {
             var itemsToUpdate = basket?.Items?.Where(x => int.Parse(x.ProductId) == productId).ToList();
 
-            if (itemsToUpdate != null)
+            if ((itemsToUpdate != null) && (itemsToUpdate.Any()))
```
jzeferino: Could be simplified to `itemsToUpdate?.Any()`

eiximenis:
Closing as is too old, and much of this has been already done :)

PR #134, opened 2017-03-23
Update script permission

BillWagner:
The mac-cli/build-bits.sh script does not have +x permissions.

I added those. There is no diff in the file, just in the git index to update the permissions on a unix or Mac system.


CESARDELATORRE:
That's good. In any case, the recommended way to build the bits is with the build container, like this:
Docker-compose -f docker-compose.ci.build.yml up

Instead of the mac-cli/build-bits.sh script that we're maintaining it in the source code just for other kind of tests or learning purposes.

BillWagner:
that makes sense. I like being able to build the bits locally (as opposed to in a container) at times.

PR #142, opened 2017-03-27
Set application discriminators, work around Redis client issue

chlowell:
A couple small changes to resolve issues I found hosting the app in Kubernetes:
1. Setting application discriminators prevents an app trying to decrypt another's cookie
1. When `Basket.API` connects to Redis, don't call `Dns.GetHostAddressesAsync` if the connection string is an IP address. It may return `[::1, 127.0.0.1]` for `"127.0.0.1"`. Then `ips.First()` is `::1`, which the Redis client can't handle.


CESARDELATORRE:
Tested. The app works good on a regular Docker Host with this PR.

PR #178, opened 2017-04-28
Fix API signatures for collections

BillWagner:
The methods that return collections should return
Task<IEnumerable<dynamic>> not Task<dynamic>


CESARDELATORRE:
Bill, I tested this PR and it breaks the app when placing the Order.

BillWagner:
Hmmm. I'll look into what I missed and submit an update when I find that problem.

PR #182, opened 2017-05-03
Run webforms with catalog microervice

BillWagner:
Retarget the changes to run the WebForms app to the dev branch.

Tests are done, and I just added a README.MD.

This is ready for review and merge.  cc @CESARDELATORRE


BillWagner:
@kallie-b @JMesser81 Can you look at the last commit, with the readme.md updates? Is that the right guidance for networking using Docker for Windows and Windows containers?

src/Web/Catalog.WebForms/readme.md
```diff
@@ -0,0 +1,80 @@
...
+  </appSettings>
+```
+
+Set that value to `false` to run with the full catalog microservice.
+
+Then, set the startup project in Visual Studio to the *Catalog.WebForms* project. Build and run, and test the application.
+
+## Configure Windows Containers
+
+The Catalog.WebForms project uses the full framework, so will only run in Windows based Docker containers. Before running in Docker, make sure you
```
kallie-b: Which "full framework" do you mean here? The full Docker framework?

Also, there have been some recent marketing changes that can make Docker versions confusing for people. For that reason, I recommend you try to consistently refer to "Docker for Windows" (which is a Docker Community Edition (Docker CE) version), as consistently as possible when referring to Docker for Windows 10. It may be nice also to link to this doc on Docker for Windows: https://docs.docker.com/docker-for-windows/install/.

src/Web/Catalog.WebForms/readme.md
```diff
@@ -0,0 +1,80 @@
...
+are running Docker with Windows containers configured. Right-click on the Docker node, and if "switch to Windows Containers" is displayed, click that. If "switch to Linux containers" is displayed, you are already running
+with Windows containers.
+
+## Update Configured IP address
+
+There is a current limitation in Docker for Windows when running with
+Windows containers where *localhost* doesn't resolve to the Docker host IP.
+Instead, you need to find the physical machine's IP address and use that for
+the external address of the services.
+
```
kallie-b: This limitation is centered around communication between a container and _its host_--we don't have the routing in place for container to be accessed (via its exposed port on the host) using localhost, but we recently added a fix so that this is possible using the container host's IP address. 
I would avoid saying this is a Docker for Windows limitation, because it's really just a Windows (HNS) limitation--but you probably don't need to say that either. If you'd like to see how I generally word this, take a look at this [StackOverflow post that I replied to](http://stackoverflow.com/questions/43769806/docker-for-windows-cannot-access-service-on-exposed-port-in-windows-container-m/43770786?noredirect=1#comment74590252_43770786). I've included descriptions of the "workarounds" for this in my answer to that post.

kallie-b:
@BillWagner I scanned the Readme for Docker for Windows sections, and I think I got them all. Let me know if you think I missed something, or if you have any questions at all! :)

BillWagner:
@kallie-b Thanks! updated based on your comments.

CESARDELATORRE:
Closing this PR as the WebForms app is not being supported here since we created the "eShopModernizing" repo for modernizing legacy apps with Windows Containers:
https://github.com/dotnet-architecture/eShopModernizing

PR #190, opened 2017-05-06
Fix mac ci build

BillWagner:
While at it, update the sh script too.

I’ll remove the WIP label once I test this on the PC


BillWagner:
Verified on the PC. It builds all projects and images.

PR #199, opened 2017-05-19
Packages update to 1.1.* and Swagger update

SychevIgor:


SychevIgor:
@CESARDELATORRE do you have a minute to look on it?

CESARDELATORRE:
@SychevIgor Checking it out now. Why did you change the following in quite a few places in this commit? : https://github.com/dotnet-architecture/eShopOnContainers/pull/199/commits/9e755046905dee5670b2a3484d163901b0c031a6
- From IAsyncRequest to IRequest
- From SendAsync() to Send()
- From IAsyncNotification to INotification
- From PublishAsync() to Publish()

jbogard:
That's the API changes from MediatR 2 to 3.

CESARDELATORRE:
@SychevIgor There's an issue with Swagger. I'm getting an Http 500 error when trying to get the SwaggerUI on the Catalog microservice, for instance, trying: http://localhost:5101/swagger/
Looks like the new version of Swagger doesn't support to use base [Route] attributes at the class level while using [HttpXXX] attributes per each method... Seems similar to this issue:  
https://github.com/domaindrivendev/Swashbuckle/issues/581

Can you check it out? I changed all the controller's methods to just use the [HttpXXX] attributes with the full path and removed all the [Route] attributes from the Catalog Web API, but still getting the same error from Swashbuckle.. I might be missing something else..

CESARDELATORRE:
I found the issue's cause. We were using the parameter "Name" of the SwaggerDoc() as a simple name, like "Catalog API v1". That was working in previous versions.
In this version it has to be a it MUST be a URI-friendly name that uniquely identifies the document, like "v1"
i.e. The following case fails with the new version of Swashbuckle:
```
services.AddSwaggerGen(options =>
{
 options.DescribeAllEnumsAsStrings();
 options.SwaggerDoc("Catalog API v1", new Swashbuckle.AspNetCore.Swagger.Info {Title = "Catalog API", Version = "v1"});
 //...
```
i.e. The following case, works ok:
```
services.AddSwaggerGen(options =>
{
 options.DescribeAllEnumsAsStrings();
 options.SwaggerDoc("v1", new Swashbuckle.AspNetCore.Swagger.Info {Title = "Catalog API", Version = "v1"});
 //...
```

CESARDELATORRE:
There are other needed changes like:
- Setting the "launchUrl" of the WebAPI as "swagger" instead of "/swagger/ui"

CESARDELATORRE:
I see the same introduced BUG in the rest of the services you modified for the new version of Swashbuckle:
`SwaggerDoc("eShopOnContainers - Ordering HTTP API", ....`
Has to be like:
`SwaggerDoc("v1", ....`

CESARDELATORRE:
@SychevIgor Please, do check the Basket Checkout action, basically when "Placing and Order" which is using the Ordering Microservice. I got some errors related to Autofac registration probably related to the type changes in MediatR. But I couldn't test this scenario enough time.
Btw, thanks a lot for the PR! - Let's finish it and we'll be able to merge it! :)

SychevIgor:
@CESARDELATORRE about  Basket http://localhost:5103/swagger/ .. I'm not sure that it worked before... because on a BascetController we have [Route("/")] attribute, it means that instead of swaggerui will be called Bascet.Get method. I can change the route value to  [Route("api/v1/[controller]")] and it will solve the problem with basket swagger file

CESARDELATORRE:
I tested the PR with the changes I told you to do so Swagger works and looks it is good to go, so I merged it. 
About the BasketController with [Route("/")] attribute, we'll fix that, but it might impact the client apps, so we'll handle that, ok?
Thanks for the PR! 👍

PR #219, opened 2017-06-22
Healthcheckcleanup

SychevIgor:
we don't need jquery.validate for such a simple page 
we are not using banners and js in healthcheck


SychevIgor:
@CESARDELATORRE ?

eiximenis:
Hi!
I merged the changes into `dev`.
Thanks!

PR #221, opened 2017-06-22
Error action extracted from CatalogController, because otherwise was …

SychevIgor:
Error action extracted from CatalogController, because otherwise was violated Single Responsibility Principal, when even to generate an Error we will create Transient ICatalogService for each error.


SychevIgor:
@CESARDELATORRE

SychevIgor:
@ardalis could you please loon on this PR? it's not merged for long time

PR #222, opened 2017-06-22
Identity in mvc fix

SychevIgor:
we are not using identityparser in AccountController. we can drop it from a constructor
2 Views survived after identity redesign and dropped only in this PR


SychevIgor:
@CESARDELATORRE

SychevIgor:
@ardalis could you please loon on this PR?

PR #223, opened 2017-06-22
DRY https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

SychevIgor:
in web.common project extracted classes copy-pasted in all api projects.
DRY https://en.wikipedia.org/wiki/Don%27t_repeat_yourself


SychevIgor:
@CESARDELATORRE

eiximenis:
Hi!
We'll take a look on this! :)

About the DRY violations, is basically because we are always dealing with (IMHO) two opposite scenarios:

1. Developing using a "microservice approach" where each microservice is "independently developed"
2. Having a single solution and a codebase mantained by one team in one single repo (which contradicts 1 :P)

For some "infrastructure" projects (i. e. EventBus) we developed it once, because it seemed obvious that in a "real" development you will have a nuget package providing this infrastructure to the rest of projects.

Not saying that this is the best option to deal with it, just the one we followed ;-)

Thx!

SychevIgor:
@eiximenis should I close this PR or rebase on top of current dev branch? I don't like when PR so long in unknown state.

CESARDELATORRE:
@SychevIgor I don't want to have "common code" across web or service projects. In a microservice approach, each microservice should be completely autonomous. Think about a different dev team that might be working on each microservice.
Now, we might use the same "building blocks" from the multiple microservices, like the Event Bus, or other libraries which are re-usable like if they were Nuget packages.
But, I don't want to have a hard dependency across all the microservices like "Microsoft.eShopOnContainers.**Infrastructure**" project or a "**Web.Common**.csproj" project which are not clear/isolated building blocks but just "DRY classes" within a common project.
Again, microservices should be completely autonomous. We don't want to have a central dependency across the different microservices unless it is something like an isolated building block (like a Nuget package).

PR #227, opened 2017-06-26
Update Startup.cs

skynode:
`builder.AddUserSecrets();` is deprecated and will not be available in ASP.NET Core 2.0.


skynode:
See [link](https://github.com/aspnet/Announcements/issues/223) for reference.

PR #232, opened 2017-07-04
StandardHttpClient setting Authentication Header

matthewDDennis:
In a couple of places the AuthenticationToken is being used instead of the AuthenticationMethod parameter.  This results in failed requests.


eiximenis:
Thanks for the PR.

Merged!

PR #241, opened 2017-07-16
Configure packages directory for restored NuGet packages

mrward:
All solutions now restore to the packages directory at the root of
the repository. This fixes packages being restored into the
src/Mobile/packages directory when the eShopOnContainers.Xamarin.sln
is opened.

https://developercommunity.visualstudio.com/content/problem/80508/references-are-broken-even-though-nugets-have-been.html


ramonTC:
Done, thx!

PR #255, opened 2017-08-03
Fix StandardHttpClient's DoPostPutAsync Method

jonorogers:
DoPostPutAsync in StandardHttpClient was locked to POST methods. Fixed to allow PUTs.


eiximenis:
Thx!
Merged!!!! :)

PR #298, opened 2017-08-29
win prefix to all windows based containers. make naming consistent

SychevIgor:
For some container we have win prefix, for some - not.  it's inconsistency

### DIFF
--- a/docker-compose-windows.yml
+++ b/docker-compose-windows.yml
@@ -59,13 +59,13 @@ services:
       - marketing.api
   
   webstatus:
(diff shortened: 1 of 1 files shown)


SychevIgor:
@CESARDELATORRE

CESARDELATORRE:
Sorry, we don't agree on this change because of the following reasons:
1. If adding a different name for Windows Containers, that should be as a TAG, not as part of the name
2. The base name of the image should be the same if we want to have MULTI-ARCH Docker images and only differentiate by different TAGs and using meta-package.

PR #309, opened 2017-09-02
Parallel dotnet build process.

SychevIgor:
it's working 2m 24s instead of 3m 48s (sequential execution)
theis only one problem... write-host not working inside the workflow
P.S. in both cases comparison was without "dotnet restore command", because it's not mandatory in .net core sdk 2.0


SychevIgor:

eiximenis:
Hi @SychevIgor 

I've merged the PR (just removed GracePeriodManager.csproj as is no longer used in netcore2)

Issue #62, opened 2017-02-26
DRY up RedisBasketRepository


ardalis:
https://github.com/dotnet/eShopOnContainers/blob/master/src/Services/Basket/Basket.API/Model/RedisBasketRepository.cs#L30

Every method starts with a call to GetDatabase(). This can be safely moved into the constructor (setting a _database class field) since this repository is being used as a Transient lifetime service. Each method would then simply reference _database instead of a local database variable.

PR #311, opened 2017-09-02
Fix "DRY up RedisBasketRepository"

egorikas:
https://github.com/dotnet-architecture/eShopOnContainers/issues/62

I made some clean-up for the RedisBasketRepository.
1) Removing `GetDatabase` from the methods
2) Inject `ConnectionMultiplexer` through constructor injection.
There are some lines in `Startup.cs`
```csharp
services.AddSingleton<ConnectionMultiplexer>(sp =>
{
	var settings = sp.GetRequiredService<IOptions<BasketSettings>>().Value;
	ConfigurationOptions configuration = ConfigurationOptions.Parse(settings.ConnectionString, true);           
	configuration.ResolveDns = true;

	return ConnectionMultiplexer.Connect(configuration);
});
```
Why don't use them? :)
3) `GetUsersAsync` -> `GetUsers`. 'cause there is no need for async after implementing constructor injection
4) Refactoring calls of `GetUsersAsync` method


ramonTC:
Thx!

PR #317, opened 2017-09-05
NoBuild missed some images

SychevIgor:
looks like this file has missed latest refactoring


SychevIgor:

PR #320, opened 2017-09-06
payment API require health check

SychevIgor:
payment API require health check
also swagger in Payment API is useless, because we don't have any API controllers here. When I had looked on Payment.API for a first time- I thought that I have some issue with it...


SychevIgor:

PR #321, opened 2017-09-06
Grace period rudement fix

SychevIgor:
super minor fix


SychevIgor:

PR #322, opened 2017-09-06
sync k8 images with local windows/linux images for consistency

SychevIgor:
sync k8 images with local windows/linux images for consistency


SychevIgor:

PR #333, opened 2017-09-15
Treat *.pfx as binary so CRLF will not change it

dvdstelt:
This might have caused an issue with the .pfx in a work we created. Was kind of weird, but this should make sure it doesn't happen again.


CESARDELATORRE:
Thanks! :)

Issue #305, opened 2017-09-01
Swagger generation


Meir017:
use the attribute `ProducesResponseType` on controllers actions.

this will make the swagger document be more accurate

SychevIgor:
I think that answer will be similar to https://github.com/dotnet-architecture/eShopOnContainers/issues/315#issuecomment-327247013 
but - it's only my assumption

CESARDELATORRE:
@SychevIgor Why is this related to Razor pages?

SychevIgor:
Only first sentence your answer:
"The focus for this project is currently most of all about the microservices, not so much about the client apps. "

Meir017:
would you accept a pull-request?

CESARDELATORRE:
@Meir017 We will, just need to test it before and we have many other priorities. It won't be long, sorry. :)

SychevIgor:
@CESARDELATORRE  looks like we can close this issue.

PR #368, opened 2017-10-09
swagger response types and code 

SychevIgor:
https://github.com/dotnet-architecture/eShopOnContainers/issues/305
@CESARDELATORRE


SychevIgor:
@CESARDELATORRE @ramonTC @eiximenis please, could you give me any feedback?

CESARDELATORRE:
OK, so you are basically adding the ProducesResponseType attribute, even when sometimes there's a single response type (like "200") but because most methods are returning an IActionResult instead of a direct DTO, right?
And sometimes, because you want to specify additional responses like "NotFound" or "Created", or your actions return IActionResult instead of a response DTO.
I think it looks good. I'm testing the PR and everything seems to work good.
@eiximenis Do you have anything else to add/comment?

CESARDELATORRE:
Merging. Thanks SychevIgor! 👍

PR #369, opened 2017-10-10
Update HealthChecks and HealthChecks.SqlServer to .NET Standard 2.0

KSYMEK:


CESARDELATORRE:
Not sure we want to upgrade this class library (Health Checks) to .NET Standard 2.0, since this is a copy of the original ALPHA code here: https://github.com/dotnet-architecture/HealthChecks
So, it is better not to diverge from there.
Also, why update to .NET Standard 2.0 if you don't need any API from any newer specific framework implementations? - I mean, if it works in .NET Standard 1.3, it'll work on more platforms/frameworks than if it works on .NET Standard 2.0. Here's the matrix:
https://github.com/dotnet/standard/blob/master/docs/versions.md

The benefit of moving to .NET Standard 2.0 is when you need APIs in newer versions while having compatibility with the rest of the frameworks. I.e. if I want to use something specific implemented in .NET Core 2.0 or UWP while .NET Standard 2.0 increased a lot the % of APIs from the multiple frameworks.
But if the APIs you are using are enough and compatible with older versions of frameworks, it is better if you target lower versions of .NET Standard as you'll have compatibility with a higher number of older versions, as shown in the URL: https://github.com/dotnet/standard/blob/master/docs/versions.md

As described in the mentioned URL/page, when choosing a .NET Standard version, you should consider these trade-off:
- The higher the version, the more APIs are available to you.
- The lower the version, the more platforms and versions implement it.

**So generally speaking, you should target the lowest version you get away with**. To inform your decision, you should consider the compatibility matrix and the APIs additions linked from the table above.

Finally, I wouldn't bother much about this Health Checks lib because we'll be moving to the officially supported library in ASP.NET 2.x, in a Nuget package, soon.. :)

Issue #362, opened 2017-10-08
[Question] Why retry policy implemented using constants instead of configuration?


SychevIgor:
Why retry policy implemented using constants instead of configuration?
For example in cold be 5, could be 6 depend from async or wait/circuit DefaultRabbitMQPersistentConnection.cs, EventBusRabbitMQ.cs, ResilientHttpClientFactory.cs

    .WaitAndRetryAsync(
                    // number of retries
                    6,

 .CircuitBreakerAsync(
                   // number of exceptions before breaking circuit
                   5,

  .WaitAndRetry(5, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)), (ex, time) =>
                {

CESARDELATORRE:
Well, I agree, we could improve this point. Once the app is deployed and working properly, usually you don't need to change those parameters, but I agree that these numbers could come from env-vars at configuration. Feel free to create a PR with it or we might improve that, eventually.

PR #373, opened 2017-10-12
retry policy defined via config

SychevIgor:
@CESARDELATORRE https://github.com/dotnet-architecture/eShopOnContainers/issues/362
I'm not sure that it's the best implementation, so I your feedback is highly appreciated


reisenberger:
_Disclosure_: I am the [Polly](https://github.com/App-vNext/Polly/) maintainer, not part of the [eShopOnContainers](https://github.com/dotnet-architecture/eShopOnContainers) team.  So really it is still for @CESARDELATORRE to comment.  But ... overall, looks reasonable to me.  

There would certainly be the opportunity to factor out this repeating code pattern, into a helper method:

```
var exceptionsAllowedBeforeBreaking = 5;
if (!string.IsNullOrEmpty(Configuration["HttpClientExceptionsAllowedBeforeBreaking"]))
{
    exceptionsAllowedBeforeBreaking = int.Parse(Configuration["HttpClientExceptionsAllowedBeforeBreaking"]);
}
```

eg `int GetConfiguration(string key, int defaultIfNotSpecified)`.  

---

Later, in Polly, we plan to add native support for configuring policies from environment configuration.  we plan to provide an `IRetryConfigurationProvider` interface, implementations could draw the configuration from various sources.  But this is not ready yet (maybe Q4 2017).  The direct approach in this PR is good in the meantime.

CESARDELATORRE:
Agreed. Looks good. I just need some time to test the PR, too. :)

CESARDELATORRE:
@SychevIgor I've tested it, so I'm gonna merge it.
However, in addition appsettings.com, it could also be provided from the env-vars at docker-compose.overide.yml and from the K8s .yml config files and the Service Fabric XML Manifests. That way you wouldn't need to re-build the Docker image if you change any value.
But, for now, this might be enough (just appsettings.json per service) since the retry policies is not something you'd be changing frequently, though... 
We can always update the docker-compose.override.yml and comparable env-var files for the orchestrators later on... This is, initially, a pretty much internal configuration, depending on the app you might not need to change it very often.

Issue #366, opened 2017-10-09
[Question] Do we have a reason to use dynamics type in OrderQueries.cs file


SychevIgor:
I just started working on https://github.com/dotnet-architecture/eShopOnContainers/issues/305 and got an interesting situation... I can't normally describe the return type for OrderController.GetOrder because type essentially dynamic.
after some research - I didn't find any reason for dynamic type here. 
May I fix the type or could you explain the idea?

@CESARDELATORRE

CESARDELATORRE:
@SychevIgor 
The reason is that being dynamic, you can just change the SQL query (new attributes/columns, etc.) and you'll get that returned. you have like a dynamic ViewModel. No need to have static classes for those ViewModels.  So, it is more agile when developing.
So, for a service used only by the same application, it is ok and more agile.

However, for a public API, it would be better to use ViewModels or "contracts".
And even for a long term application, I find that having explicit ViewModel classes is safer and also, as you mentioned, better for Swagger or even other "unknown clients".
So, after testing that PR well, we can merge it. But, can you test it with all the client apps? Including the SPA app and the Xamarin app?

CESARDELATORRE:
Closing this issue. We decided to accept the PR and use predefined ViewModel DTOs classes as it is better for public APIs but also for long term microservices, even if they are just used by the same application.
Also, as mentioned by SychevIgor, if you want to specify response types for swagger, you need to use explicit DTO classes as the return type. Therefore, predefined DTO classes allows you to offer a richer information from Swagger. That will benefit the API documentation and compatibility when consuming an API.

PR #380, opened 2017-10-14
no more dynamics as a return types from controllers in Ordering app

SychevIgor:
https://github.com/dotnet-architecture/eShopOnContainers/issues/366
Without it- we can't specify response types for swagger.
also dynamic types in this case- have no benefits

@CESARDELATORRE @ramonTC @eiximenis

### DIFF
--- a/src/Services/Ordering/Ordering.API/Application/Queries/IOrderQueries.cs
+++ b/src/Services/Ordering/Ordering.API/Application/Queries/IOrderQueries.cs
@@ -5,10 +5,10 @@
 
     public interface IOrderQueries
     {
-        Task<dynamic> GetOrderAsync(int id);
+        Task<Order> GetOrderAsync(int id);
 
-        Task<IEnumerable<dynamic>> GetOrdersAsync();
+        Task<IEnumerable<OrderSummary>> GetOrdersAsync();
 
-        Task<IEnumerable<dynamic>> GetCardTypesAsync();
+        Task<IEnumerable<CardType>> GetCardTypesAsync();
     }
 }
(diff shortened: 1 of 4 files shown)


CESARDELATORRE:
@SychevIgor We were using "dynamic" here so it is more agile if changing/adding attributes in the SQL queries in Dapper it will just be returned without needing to change any ViewModel or DTO classes for that. That is good for a single and auto-contained monolithic app. Not so good for a microservice based application where explicit "service-contracts" are more important. 
I see PROS and CONS in both approaches. We need to decide about it.
In any case, when using specific ViewModel types, have you tested it with ALL the client apps, SPA and Xamarin apps in addition to the MVC app?

SychevIgor:
@CESARDELATORRE I did my tests with swagger ui only. My test methodology was: make a request from swagger UI, copy-paste json to jsontocshare converter and copy the result to the project. I can guarantee that json output before (with dynamic) is the same as with new strongly typed.

I did it primary, because swagger can't show us the contract(return model) before the first call as a result - you can't generate proxy based on swagger metadata. If we can't show the response model - it means that we are not using half of goods from swagger .

SychevIgor:
Upps... I missed the fact that- MVC/SPA apps are working with updated ordering microservice without issues. It was tested. I didn't test  Xamarin, because I'm not focusing my attention on mobile part of this reference architecture

CESARDELATORRE:
@SychevIgor 
I tested this PR with the MVC, SPA and the **Xamarin mobile app** consuming the microservices/containers.
I think that in the long term it is better to have explicit ViewModel classes (like contracts) instead dynamic ViewModels, so I merged this proposal. 
Thanks! :)

PR #381, opened 2017-10-14
last .net core 2.0 part

SychevIgor:
all projects finally migrated to .net standard 2.0
@CESARDELATORRE @ramonTC @eiximenis


CESARDELATORRE:
@SychevIgor I might be wrong, but looks like after merging these changes, the Order process was failing. Not sure if it was the code or my environment, but after reverting the code, the application order process is working properly...
I cannot test it better due to lack of time today. 
If possible, can you test/confirm if this is working right?
Thanks a lot for your PRs! 👍

SychevIgor:
@CESARDELATORRE I will test it again today evening... But I remember that with this commit everything works on my machine (login via mvc/spa apps, add goods to a busket, submit the order).

PR #410, opened 2017-11-20
delete config map fix

alexandervantrijffel:


ramonTC:
The command ExecKube -cmd 'delete configmap config-files' is already set at:
https://github.com/alexandervantrijffel/eShopOnContainers/blob/7b54c05b7182422404831327ceed26a0b13e0522/k8s/deploy.ps1#L103

We will remove it from the comment "# start sql, rabbitmq, frontend deploymentsExecKube -cmd 'delete configmap config-files'" as it provokes confusion.

Thx!

PR #417, opened 2017-11-27
Converts to use multi-stage build Dockerfile

yanchenw:
1. Changes the build context of all services to the root level to be
able to access the dependencies for each project.
2. Commented out the prepublish scripts, these steps are moved into the
Dockerfile

Did not change any script yet, need help to identify what scripts we
need to change.


eiximenis:
Thanks @yanchenw.

I'll take a look on this!! :)

CESARDELATORRE:
We'll do some testing with it in this branch features/docker-multi-arch
Thanks @yanchenw !! 👍

Issue #437, opened 2017-12-13
[Feature Request] SPA update to Angular6


SychevIgor:
I think we can update angular from 4 we are using now to 5.1 during next release cycle https://github.com/angular/angular/releases 
What do you think about it? @CESARDELATORRE @ramonTC

thiagojsvix:
OK

Em 13/12/2017 08:12, "SychevIgor" <notifications@github.com> escreveu:

ramonTC:
It is fine.
I've checked out the breaking changes which introduce ng v5 and it shouldn't be a big deal

SychevIgor:
https://github.com/dotnet-architecture/eShopOnContainers/pull/441

PR #441, opened 2017-12-14
angular5

SychevIgor:
first version of spa ported to angular5. we have some warning and I'm not sure about work in somebody's else environment... but at least it's working on my machine.

@CESARDELATORRE  @ramonTC


SychevIgor:
https://github.com/dotnet-architecture/eShopOnContainers/issues/437

ramonTC:
I got problems with this new version as it breaks some UI components:
image
image

I've installed dependencies and generated the spa scripts:
npm install
npm run build:dev

And Run eShopOnContainers from vs2017.

SychevIgor:
@ramonTC interesting, because I didn't find this issue on my machine. If I will have time- I will test and will also try to investigate it

PR #453, opened 2018-01-04
Optimize multi-stage build Dockerfile

yanchenw:
Remove explicitly copied dependencies, copy entire solution
into the conainter and depends on the cache to reduce build
time.

Addd temporary solution file which does not contains the
dcproj file to work around the issue that the dotnet cli
cannot handle dcproj.


CESARDELATORRE:
Hey! - Thanks for this improvement! 👍 
However, I don't like to have yet another solution (the one you added named **eShopOnContainers-ServicesAndWebApps-MultiStageBuild.sln**), because whenever we add anything new into the VS solution we must update the same thing into this other solution. Is this mandatory/needed?
If it is just because of a warning, I prefer to have the warning than having to maintain a new additional solution file.
Thoughts?

CESARDELATORRE:
Another question, what these parameters do within the dockerfile?
-nowarn:msb3202,nu1503

CESARDELATORRE:
Can you confirm if we really need the eShopOnContainers-ServicesAndWebApps-MultiStageBuild.sln solution file? - I'm ok if we just have a warning when using the other .sln file because dotnet cli
cannot handle dcproj files.

yanchenw:
@CESARDELATORRE  The new solution file is required because the dcproj file will cause the build to fail, not just a warning. But it can be removed once we have Docker SDK files integrated into dotnet SDK. I will start to work on that and keep you updated about this.

yanchenw:
@CESARDELATORRE  -nowarn:msb3202,nu1503 is to now show some project not found messages when doing restore.

CESARDELATORRE:
@yanchenw I see.. in dockerignore you are ignoring (not copying) all the .sln files but this one, that .sln is the one being used by default when running dotnet publish, right?

dockerignore:
...
docker-compose.dcproj
*.sln
!eShopOnContainers-ServicesAndWebApps-MultiStageBuild.sln
...

OK, we'll change this once we have Docker SDK files integrated into dotnet SDK..

PR #465, opened 2018-01-16
Replace Autofac with TinyIoC in the mobile app

### LINKED ISSUE #393: Memory Leak problem on Xamarin application
As I check the source code, there is a significiant memory leak problem with ViewModels.

1. When you need to a view, it's created using Activator.CreateInstance (here: https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/src/Mobile/eShopOnContainers/eShopOnContainers.Core/Services/Navigation/NavigationService.cs#L113).

That's OK.

2. When a view is created, corresponding ViewModel is automatically created using _container.Resolve (here: https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/src/Mobile/eShopOnContainers/eShopOnContainers.Core/ViewModels/Base/ViewModelLocator.cs#L113).

The problem is here. ViewModels are transient. Every time you Resolve a ViewModel, you get a new instance. But you never release/dispose a ViewModel and it's dependencies. Since you are resolving it from the root container, it will cause memory leak by time (see autofac, it explicitly declares to avoid resolving from the root container: http://autofaccn.readthedocs.io/en/latest/lifetime/index.html).

davidbritch:
Autofac isn't massively performant in Xamarin apps, typically taking ~210ms to register a type, and ~290ms to resolve a type.

TinyIoC typically takes ~6ms to register a type, and ~50ms to resolve a type. This has a noticeable effect on app performance, particularly on Android where the startup time has been significantly reduced.

As well as performance, replacing Autofac with TinyIoC has a number of other advantages:

- The unit tests run more quickly.
- The unit tests are more stable.
- It's solved a huge memory leak caused by how we were using Autofac - fixes #393


jsuarezruiz:
Nice work!

muhaym:
Any update on startup time measurements before and after the changes?

Kapusch:
This is funny to discover this merge replacing Autofac by TinyIoC, because I'm currently doing the right opposite !

Indeed, I was glad of TinyIoC until I've starting to mock my services with Moq library.
Whatever I've tried for handling this, I've always got a PlatformNotSupportedException error.

And now I've read @davidbritch's comment, I hope performances won't be as much impacted...

muhaym:
@Kapusch Trust me performance with AutoFac is miserable in Android. find something else to mock your services

Kapusch:
Indeed @muhaym , I've finally reverted back to TinyIoC for the exact same reason : PlatformNotSupportedException.

I've had a hard time on this, because I really like Moq library and was thinking to be able using it for UI Tests, but no.

At the end and as you've suggested, I've implemented mock classes instead.

PR #472, opened 2018-01-18
Add deployment script for Mac/Linux

karolz-ms:
The script has some limitations compared to the PowerShell version--see comments in the script. @CESARDELATORRE thought it is a useful starting point regardless.


CESARDELATORRE:
@karolz-ms Thanks! 👍

PR #473, opened 2018-01-19
Fix about some async calls.

PMExtra:
The `DbSet<T>.Any()` is a database I/O operation, it's should be async.
But the `DbSet<T>.AddAsync<T>(T entity)` is only used to allow special value generators, in the document:

> This method is async only to allow special value generators, such as the one used by  'Microsoft.EntityFrameworkCore.Metadata.SqlServerValueGenerationStrategy.SequenceHiLo', to access the database asynchronously. For all other cases the non async method should be used.

So I think it should be non async.


SychevIgor:
Sound 100% logical. 
@CESARDELATORRE

eiximenis:
LGTM
merged!
Thx

PR #474, opened 2018-01-20
Update MediatorModule.cs

EnricoTirotta:
Simple update of interfaces name in comments


SychevIgor:
Agree, I didn't change when migrate it from old nuget package.
@CESARDELATORRE  could you merge it?

mvelosop:
Thanks!

PR #487, opened 2018-01-29
Update AI k8s package version, otherwise kubernetes deployment will fail

yantang-msft:
Kubernetes deployment will fail because of this issue: https://github.com/Microsoft/ApplicationInsights-Kubernetes/issues/73, which is fixed in version 1.0-beta3


yantang-msft:
@CESARDELATORRE Could you please take a look at this PR?
Without this change, K8s deployment doesn't work at all.

CESARDELATORRE:
@yantang-msft  - Thanks for the heads-up. I'm merging this PR. 
Btw, Microsoft.ApplicationInsights.Kubernetes is already in BETA5:
https://www.nuget.org/packages/Microsoft.ApplicationInsights.Kubernetes/1.0.0-beta5 

Have you tried/tested BETA5 version with eShopOnContainers?
Thanks,

yantang-msft:
@CESARDELATORRE Not yet.
The description for beta5 is missing. Beta4 just added some more properties into the telemetry, not a crucial change.

PR #491, opened 2018-01-30
Changed AddRange to AddRangeAsync

raboud:
in all the Context Seeding methods we call AddRange instead of AddRangeAsync even though the methods themselves are marked async.


mvelosop:
Thanks!

PR #493, opened 2018-01-30
Consolidated redundant code

raboud:
there was redundant code in each service so moved to a couple of new common libraries


eiximenis:
Hi!
Many thanks for your update!

@CESARDELATORRE  @SychevIgor  What is your feeling on that? Let just me know if it is ok for you to merge this **BUT DO NOT PERFORM THE MERGE!**

Currently we are in the middle of a big refactor (in a branch called `bff`) so to avoid merge problems I prefer to defer the merge of this PR onto dev until `bff` development is finished. This PR and the current refactor in bff branch do not affect the same code, but some files (sln, some csprojs) are impacted by both.

Thanks again.

CESARDELATORRE:
@raboud  
I'm not sure about this change. Initially, I prefer not to have "Common" libraries between microservices which are not really specific artifacts with a very clearly defined boundary like the building blocks or NuGet packages.
DRY (DON'T REPEAT YOURSELF) is good/important within a single project or a single autonomous boundary. In this case, a single microservice.
But in a microservice architecture it is more important to have independent/autonomous microservices that are not dependent from "common" code so they can be developed independently by different development teams without being impacted by changes from the other team.
The only shared code/libraries should be really independent code like "Building Blocks" or Nuget packages, i.e.: Serializers, Event-Bus implementation, HttpResilient implementation, etc. at the end of the day, things that you'd put in a stable NuGet package to be shared across applications.But I don't like to have a "Common" library just for the sake of not having some redundant code in microservices.
It is more important to respect their autonomous development.

Do you really think this code is a specific building block with a very clear functional boundary?
Or is it just a couple of methods/middleware that it happens to be used in several microservices?

I think it is more about the second case. So, I'd initially wouldn't create this "Common" library.
But this is a good point of discussion on microservices so I'd really like to get your points of view! :)

Thoughts?
Cesar.

SychevIgor:
https://github.com/dotnet-architecture/eShopOnContainers/pull/223 I did similar request for filters and exception before. I was agree on common/shared libraries before and still would like to have one as less as possible copy-pasted code.

raboud:
@CESARDELATORRE  I understand your point about independence....that being said if all the API micrososervices are being setup the same way and it was important enough to do it the same way on each microservice api ... it seems to be a project level building block. And while each microservice is independent they exist to work together to each do a portion of a much bigger task. 

I have no dog in this fight. I am basing a big project loosely on this and as such I wanted to try and pay some back.

Feel free to close this PR at your discretion.

CESARDELATORRE:
Don’t make me wrong, we love these discussions. It is precisely why many developers might think. And really, thanks for taking the time for the PR.
My point is that if we define a clear boundary and feature purpose for the building block, I’d agree on having it.
What I don’t want to have is a “laundry laundry list” of common code across microservices, because that would be an anti-patttern in a microservices architecture.

What unique feature’s name would you call to this new building block?

Thanks! :)
Cesar.

EnricoTirotta:
It's a very good discussion. **DRY principle speaks about "knowledge"**, "Every piece of knowledge", it's not about "patterns" or "code". I forget often. Like @raboud, i everyday have the temptation to cut code and avoid code duplication because we, as developers, feel a good sensation in less code writing... 😄 Less is more... but it's not always true! With DDD in mind, i'm changing my approach: i use generics, shared kernels, shared libraries with much more attention. I always have a big danger alarm in my mind if i try to use them because I have no chances for future decisions, i link my code.

I think that code with real-DRY principle in mind is to code with a specific importance on knowledge management and consistency in front of the DRY-cut-code principle that introduce **unnecessary complexity**. DRY, if we see it only from code point of view, put together **pieces of code completely extraneous and hard coupling them**. So, repeat it is not the evil.

Obviously it's just my point of view.

Have a nice day,
Enrico

mvelosop:
Hi, a bit late to the discussion and I think it's a very good one, so I'd like to drop in my two cents!

I more in line with @raboud's point of view, but having said that, I'd like to present my view of this issue in the form of a question: 

Say we agree on a set of common "building blocks" and turn them into a NuGet package that we use in several microservices.

Why is this different than using say Autofac or MediatR when a) I can use different versions or b) include the source to handle a that special need for a specific project, on an exception basis?

This allows you to have the best of both worlds, DRY as much as possible, CUT & PASTE when you really need it.

mvelosop:
Closing this PR, as the conversation stalled.

Besides, it looks that this type of "DRYing" seems to be more of a con than a pro when applying DDD!

PR #498, opened 2018-02-01
BasketItem - ProductID is a string can cause a formatexception in int.parse

raboud:
During at least one unit test, we get a basket with an item who's product it is 'productId' not a number.

when a product price update event is sent to the basket service... there is a select that tries int.Parse on the basket item.ProductId thus throwing an exception.

Gut fee; would be to change the BasketItem.ProductID to an int to match everywhere else...gut feel this was done on purpose to show the Independence  of the Basket.API from the other APIs. 

So i changed the product price update event to select on a string instead of an int...so it wont exception.


raboud:
woops too many commits

PR #499, opened 2018-02-01
Parse error

raboud:
During at least one unit test, we get a basket with an item who's product it is 'productId' not a number.

when a product price update event is sent to the basket service... there is a select that tries int.Parse on the basket item.ProductId thus throwing an exception.

Gut fee; would be to change the BasketItem.ProductID to an int to match everywhere else...gut feel this was done on purpose to show the Independence  of the Basket.API from the other APIs. 

So i changed the product price update event to select on a string instead of an int...so it wont exception.


raboud:
@mvelosop 

I did it in 2 steps as I did not want to .ToString() for each member in the DB just the one single time.

mvelosop:
Yeah, you're right, then it would just be fixing the indentation with spaces instead of tabs.

raboud:
@mvelosop 
okay color me stupid. I have never modified a pull request after submitting before. So how do I do that.

CESARDELATORRE:
@raboud - You just need to change/push the code/content in your GitHub repo and the PR will take it into account.

raboud:
@CESARDELATORRE thanks
@mvelosop tabs converted to spaces

mvelosop:
Thanks @raboud!

PR #502, opened 2018-02-06
Add reference packages to TestRunner app so it builds successfully

kdubau:
Ever since `src/Mobile/eShopOnContainers/eShopOnContainers.Core/eShopOnContainers.csproj` was moved to .NETStandard (and SDK style .csproj), the eShopOnContainers.TestRunner.iOS.csproj hasn't been compiling in `Debug|iPhone` and `Release|iPhone` configurations. 

Turns out, with the .NETStandard core library, these target/head projects need these references as well to produce a successful build.


kdubau:
It's worth noting the `eShopOnContainers.TestRunner.iOS.csproj` project had these references before the switch to .NETStandard, so perhaps they just got lost in the jumble.

https://github.com/dotnet-architecture/eShopOnContainers/blob/bee76a45239693e2a3f21128ba4b0ad8511057ed/src/Mobile/eShopOnContainers/eShopOnContainers.TestRunner.iOS/eShopOnContainers.TestRunner.iOS.csproj

Issue #508, opened 2018-02-10
Guid Bug in EventBusService.cs


Eneuman:
Azure Service Bus will detect all messages as duplicates (if duplicate detection is enabled on the topic subscription).
The following code will always generate a MessageId of "00000000-0000-0000-0000-000000000000".

https://github.com/dotnet-architecture/eShopOnContainers/blob/200e36ee56926ca359791ef83592ce75cddb042f/src/BuildingBlocks/EventBus/EventBusServiceBus/EventBusServiceBus.cs#L49

Eneuman:
PR #509 fixed this issue so I'm closing it.

PR #509, opened 2018-02-10
Changed how the MessageId is created.

Eneuman:
This code change will resolve a duplication detection problem in Azure Service Bus.
Azure Service Bus uses the MessageId to dected if a message is a duplicate.
The current code always generates a Guid looking like this: "00000000-0000-0000-0000-000000000000".

This PR fixes the following issue: https://github.com/dotnet-architecture/eShopOnContainers/issues/508

Maby it's even better to set MessageId to 
`@event.MessageId.ToString()` ?


eiximenis:
Thanks!!!!!

PR #511, opened 2018-02-12
Changed message to use the already converted body variable.

Eneuman:
The body variable was not being used.


Eneuman:
I'll split this into two seperate PRs

PR #512, opened 2018-02-12
Removed duplicate string to byte conversion

Eneuman:
The "Body" property now uses the already converted "body" variable.
Before, the "body" variable was not used at all.


eiximenis:
Thx!!!!

PR #519, opened 2018-02-22
Add ProductInfoController

kbilsted:
Every service should have an enpoint detailing its version. Great for bugreporting and monitoring the state of a deployment

### DIFF
--- a/src/Services/Catalog/Catalog.API/Controllers/ProductInfoController.cs
+++ b/src/Services/Catalog/Catalog.API/Controllers/ProductInfoController.cs
@@ -1,0 +1,21 @@
+using System.Diagnostics;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace  Microsoft.eShopOnContainers.Services.Catalog.API.Controllers
+{
+    [Route("api/v1/[controller]")]
+    public class ProductInfoController : Controller
+    {
+        [HttpGet]
+        public async Task<IActionResult> Get()
+        {
+            string assemblyVersion = Assembly.GetExecutingAssembly().GetName().Version.ToString();
+            string fileVersion = FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location).FileVersion;
+            string productVersion = FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location).ProductVersion;
+
+            return Ok($"assemblyVersion: {assemblyVersion }, fileVersion: {fileVersion}, productVersion: {productVersion}");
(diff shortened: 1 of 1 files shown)


kbilsted:
notice that it returns a string rather than a structured json document. Currently the PR is meant as an outset for a discussion rather than the finished product

SychevIgor:
Do we have a consumer of this API?

kbilsted:
I can easily name 3 

1. A monitoring dashboard showing what is deployed in each environment

2. Testers filing a bug or cannot verify a bugfix- perhaps the fix wasn't even deployed

3. Customers filing bugs- it's always nice to know the code they were running on in order to reproduce

SychevIgor:
I'm not an owner of this project, but if API not consumable - there is zero sense to add it. In your PR is just one file modified -> nobody consumed now

kbilsted:
The whole idea of loose coupling is that you don't need to know your consumer(s).

eiximenis:
@CESARDELATORRE thoughts?

IMHO this PR address one good scenario (reporting version info) but the overall solution goes far that has been proposed here.

mvelosop:
Hi @kbilsted, @eiximenis,  I think it's a great idea to have this sort of information.

I would probably prefer this implemented as a middleware instead of a controller for each application, but that's just an implementation detail.

As for the use of the the information, just as you mentioned, it could be displayed on the web pages, so users could use it to report back any issues. That's something VERY valuable for production applications.

It could also be recorded in a startup log trace as this one:

image

Which is currently implemented in WebStatus: 

https://github.com/dotnet-architecture/eShopOnContainers/blob/3bdf3eb8736dd4c9b8e34ab78b5ca4450e025560/src/Web/WebStatus/Program.cs#L104-L124

The main issue I see to implementing the versioning info in this repo is setting the version number for the microservices.

I think the best way would probably be taking the information from the relevant git repo values (tags, commits), as explained in this article: [Access Git Information from MSBuild and Code](http://www.cazzulino.com/git-info-from-msbuild-and-code.html), but that wouldn't make much sense in eShopOnContainers, since it's a mono-repo.

@nishanil and me actually talked about this subject a week ago, but even though we both agreed this was a good idea and important for production apps, it didn't seem to make much sense to implement in eShopOnContainers, for it would be a bit meaningless because of the mono-repo thing.

Thoughts?

sughosneo:
Hi Everyone, this PR has been stale for a long time now. And some thought has already been shared by @mvelosop to mention why it might not be a good fit for `eShopOnContainers` repo.
So I am closing this PR as of now. If we need to discuss more on this topic, it would be better to open a fresh issue/PR. 

Thank you.

Issue #497, opened 2018-02-01
Can not run all test together.


raboud:
when i select all test and run Xunit aborts part way thru.

found the first issue

ProductPriceChangedIntegrationEventHandler.UpdatePriceInBasketItems
is in the int.Parse that one is a pretty easy fix

the other one seems to be cause by

Send_Checkout_basket_and_response_ok_status_code

[2/1/2018 10:19:21 AM Warning] Ignoring the load test contained in file 'f:\code\github\eshoponcontainers\test\services\loadtest\basket.loadtest' as it is not supported in Unit Test Explorer. To run this test, open the load test file and run it from its editor. For more information see: http://go.microsoft.com/fwlink/?LinkID=236838&clcid=0x409
[2/1/2018 10:19:21 AM Warning] Ignoring the load test contained in file 'f:\code\github\eshoponcontainers\test\services\loadtest\campaigns.loadtest' as it is not supported in Unit Test Explorer. To run this test, open the load test file and run it from its editor. For more information see: http://go.microsoft.com/fwlink/?LinkID=236838&clcid=0x409
[2/1/2018 10:19:21 AM Warning] Ignoring the load test contained in file 'f:\code\github\eshoponcontainers\test\services\loadtest\catalog.loadtest' as it is not supported in Unit Test Explorer. To run this test, open the load test file and run it from its editor. For more information see: http://go.microsoft.com/fwlink/?LinkID=236838&clcid=0x409
[2/1/2018 10:19:21 AM Warning] Ignoring the load test contained in file 'f:\code\github\eshoponcontainers\test\services\loadtest\locations.loadtest' as it is not supported in Unit Test Explorer. To run this test, open the load test file and run it from its editor. For more information see: http://go.microsoft.com/fwlink/?LinkID=236838&clcid=0x409
[2/1/2018 10:19:21 AM Warning] Ignoring the load test contained in file 'f:\code\github\eshoponcontainers\test\services\loadtest\orderproducts.loadtest' as it is not supported in Unit Test Explorer. To run this test, open the load test file and run it from its editor. For more information see: http://go.microsoft.com/fwlink/?LinkID=236838&clcid=0x409
[2/1/2018 10:19:21 AM Informational] ------ Run test started ------
[2/1/2018 10:19:24 AM Informational] [xUnit.net 00:00:01.2349836] Received 3 results from 3 requests
[...]

PR #526, opened 2018-03-01
Fix aborting integration tests

mvelosop:
Fixed dummy basket data so the CreateOrderCommand (created from basket data) passes validations.


mvelosop:
As reported by @raboud on issue #497.

@CESARDELATORRE, @unaizorrilla this fixes the reported problem, but tests should just fail because of a validation failure, not abort. I will keep investigating tomorrow, but would appreciate any suggestion/orientation. 

It looks like the failure occurs on the MediatR behaviours pipeline.

TIA.

PR #529, opened 2018-03-03
Create order

raboud:
#497 pointed to the fact the when validation errors occurred an unhandled exception occurred.

This PR catches the exception and returns a default result to indicate a failure.

While hunting the down, I noticed that the IntegrationTest assembly was misnamed FunctionalTest so I corrected that as well - Seemed way to minor for its own PR


mvelosop:
Thanks @raboud

PR #532, opened 2018-03-04
make servicebus deployment more accurate and self describable

SychevIgor:
I spend about an hour recognizing that I need key from topics, not from namespace
This fix maybe will save sometime to others.


SychevIgor:
@christian147 @eiximenis @ramonTC

PR #536, opened 2018-03-04
drop rudiments from identity api

SychevIgor:


SychevIgor:
@ mvelosop maybe you also can look on this PR?

SychevIgor:
Any feedback?

CESARDELATORRE:
If we were going to use 2FA I guess we'd need that code, but we haven't really implemented it in eShopOnContainers...
I think I agree, please @mvelosop test that nothing breaks when removing the code from this PR (I took a look to the code and it seems safe) and then merge, when possible, ok?

Thanks @SychevIgor 👍

mvelosop:
Hi @SychevIgor, I'm sorry I missed this completely, as @CESARDELATORRE says, this should be innocuous, will test anyway!

mvelosop:
Everything seems ok! Thanks @SychevIgor.

PR #540, opened 2018-03-09
Concurrency fix for EventBus

KoLogicDev:
Prevent a possible race condition in InMemoryEventBusSubscriptionsManager


Alexbits:
There is no need for ```handler``` field. To fix an event invocation and make it thread safe, it's enough to change it to

```csharp
   private void RaiseOnEventRemoved(string eventName)
        {
            OnEventRemoved?.Invoke(this, eventName);
        }
```

PR #554, opened 2018-03-23
Remove private setters to make class immutable

Alexbits:
Remove private setters to make a class truly immutable. Otherwise, properties can be updated outside of constructor. This also makes code cleaner. Same in docs https://github.com/dotnet/docs/pull/4755


mvelosop:
Thanks @Alexbits!

PR #551, opened 2018-03-23
Make Address value object truly immutable

Alexbits:
It's better to remove private property setters to make Address value object truly immutable. Otherwise, it's possible to change the object by setting properties inside its methods for example.

``` 
protected override IEnumerable<object> GetAtomicValues()
    {
       Street = "Blah blah blah";
       City = "Blah blah blah";

        // Using a yield return statement to return each element one at a time
        yield return Street;
        yield return City;
        yield return State;
        yield return Country;
        yield return ZipCode;
    }
```

PR #556, opened 2018-03-23
Improve consistency of Entity class

Alexbits:
IMHO. The Entity class should not expose the list of domain events as ```List<INotification>```. This allows to modify the list of events without accessing Add/Remove methods. I propose to return ```IReadOnlyCollection<INotification>``` instead and add ```ClearDomainEvents()``` method to clear events.


CESARDELATORRE:
I initially agree on the suggestion/improvement. That way it would be better encapsulated, a more DDD (Domain-Driven Design) approach. The entity Behavior related to the events would be better controlled than just by making public the domain event List. I agree. 
We'll review/test it when possible and we'll come back to this PR and discussion, when possible, ok?
Thanks a lot for your contribution.

Alexbits:
Great, thanks @CESARDELATORRE

mvelosop:
Hi @Alexbits, I'm sorry I skipped your comment on #551 about this PR.

Yes, please create the new PR as in #560 but in a branch OTHER THAN **dev**.

Closing this waiting for the new one.

Thanks for your PR!

PR #560, opened 2018-03-26
Make Address value object truly immutable

Alexbits:
It's better to remove private property setters to make Address value object truly immutable. Otherwise, it's possible to change the object by setting properties inside its methods for example. Same in the docs dotnet/docs#4742

protected override IEnumerable<object> GetAtomicValues()
    {
       Street = "Blah blah blah";
       City = "Blah blah blah";

        // Using a yield return statement to return each element one at a time
        yield return Street;
        yield return City;
        yield return State;
        yield return Country;
        yield return ZipCode;
    }

### DIFF
--- a/src/Services/Ordering/Ordering.Domain/AggregatesModel/OrderAggregate/Address.cs
+++ b/src/Services/Ordering/Ordering.Domain/AggregatesModel/OrderAggregate/Address.cs
@@ -6,11 +6,11 @@ namespace Microsoft.eShopOnContainers.Services.Ordering.Domain.AggregatesModel.O
 {
     public class Address : ValueObject
     {
-        public String Street { get; private set; }
-        public String City { get; private set; }
-        public String State { get; private set; }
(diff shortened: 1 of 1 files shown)


mvelosop:
Thanks @Alexbits, 

Merged the PR this time, but it's a better practice to make changes in a branch other than **dev** in your repo, when submitting the PR, it will go back to dev in the main repo, so there's no need to make changes in your **dev** branch.

Best regards!

mvelosop:
Hi @Alexbits,

Just wanted to let you know the we recently discovered that this change, as well justified as it is, creates an unexpected bug, as the json deserializer just can't assign values to read-only properties and this causes the address information to get lost while going through the messaging service (RabbitMQ).

So it will have to be reversed.

You can check it yourself, just place an order and it'll have no address. Then change the properties to private set and it'll work fine.

You might just as well feel like making the PR yourself, I'll do it otherwise.

Best.

cc: @CESARDELATORRE

PR #561, opened 2018-03-26
Improved encapsulation of DomainEvents collection in Entity class

Alexbits:
This PR is a replacement for #556 

The Entity class should not expose the list of domain events as List<INotification>. This allows to modify the list of events without accessing Add/Remove methods. I propose to return IReadOnlyCollection<INotification> instead and add ClearDomainEvents() method to clear events.

### DIFF
--- a/src/Services/Ordering/Ordering.Domain/SeedWork/Entity.cs
+++ b/src/Services/Ordering/Ordering.Domain/SeedWork/Entity.cs
@@ -21,17 +21,22 @@
         }
 
         private List<INotification> _domainEvents;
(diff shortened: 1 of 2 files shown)


CESARDELATORRE:
Yes, I right this approach better. It is better encapsulated, more DDD, encapsulating Domain Entity's behavior within the entity.
Let's test it well and merge when ready. 👍

mvelosop:
Thanks @Alexbits!

Alexbits:
Thanks @CESARDELATORRE @mvelosop!

PR #571, opened 2018-03-29
Catalog. Fixed async/await misuses

Marusyk:
Fixed some async/await misuses. Please review. Thank you in advance


CESARDELATORRE:
@Marusyk - Thanks for these granular PRs. I agree on most of the changes related to C# conventions/style. We'll be testing it and merging them when possible. We're in the middle of some important merges so we'll merge yours when impact is lower.
Thanks for your collaboration! 👍

Issue #586, opened 2018-04-21
Swagger Auth Failing


darnoldi:
Hi there,

Auth is working fine through the MVC app, can login and view catalog and marketing.

When trying to auth through the Swagger UI on the marketing, location or ordering API's I get
`Sorry, there was an error : unauthorized_client`

In swagger I select the 'Authorize' button and in the popup I leave all at default and submit but get the error. Even If I check the scopes button I get the error?

What am I doing wrong?

Windows 10,
Docker CE 18.03.0
VS 2017 15.7.0 Preview 4

Thanks

VladislavAntonyuk:
you have to paste your token in swagger auth modal window

darnoldi:
> you have to paste your token in swagger auth modal window

Thanks, doesn't clicking authorize in swagger go to the authserver and request a token?

![swagger1](https://user-images.githubusercontent.com/7987624/39120669-5ff04392-46ef-11e8-896b-19162824f07b.png)
![swagger2](https://user-images.githubusercontent.com/7987624/39120670-60264140-46ef-11e8-93fc-1ad0d373f99e.png)
![swagger3](https://user-images.githubusercontent.com/7987624/39120672-6054e022-46ef-11e8-9649-7f7e2ea10813.png)
If I click the authorize button in swagger (image 1), the window is pre-polulated with values which I think should request the token from the auth server (image 2). but accepting those values and clicking authorize again gives me the error message in (image 3)

Thanks for your help

darnoldi:
> you paste incorrect client_id

I'm not pasting any id, that is the one that is pre-populated. Which should I use?

Thanks

VladislavAntonyuk:
You have to pass you client id which is stored in DB

darnoldi:
The client ids are created in Identity/config.cs aren't they?

` new Client
                {
                    ClientId = "marketingswaggerui",
                    ClientName = "Marketing Swagger UI",
                    AllowedGrantTypes = GrantTypes.Implicit,
                    AllowAccessTokensViaBrowser = true,

                    RedirectUris = { $"{clientsUrl["MarketingApi"]}/swagger/o2c.html" },
                    PostLogoutRedirectUris = { $"{clientsUrl["MarketingApi"]}/swagger/" },

                    AllowedScopes =
                    {
                        "marketing"
                    }
                },`

VladislavAntonyuk:
you have to pass your user client id

darnoldi:
[...]

Issue #562, opened 2018-03-28
Invalid_request on auth from Swagger for Location API


NicolasReyDotNet:
Hi,

When trying to authorize Swagger to grant an access to Location API, I get redirected with an Invalid_request error from Identity API.

[Auth from swagger](https://ibb.co/mh6HVS)
[Resulting error](https://ibb.co/g3YLjn)

Full URL is the following
http://localhost:5105/home/error?errorId=CfDJ8HlsH9uoqgBKgA8HZFcuGwLDP4TFpXOVCvyC4iWl4UbfKv3yotq6aSilta9K8IPgrbrbZvUQivH7zw8ln4hpassJYzpR8P0J65QyDcA4_Nv1xVB1PF4V_m6rs1A8YmXpTUrwp_EsQiCypEY8IXM_Y_bIdZ-4ZpPfbqjOhOqshDWWlm0mvDvs0w9XIblWsPqJDEpxPuYvsLTWsXEZJonQVvON3ke9P7gZpto-4wR1yjXi35y9oe4AMPW-Rkj1RY4ItBVY-YN4-w82LcDqNcRqCWBDhZRBLxC3SKKK2Vi8lC2E

mvelosop:
Hi @NicolasReyDotNet, could you check quick fix on issue #586?

I was having the same problem as you, only the record that has to changed for Location.Api is the one with port 5109, so the complete RedirectUri should be: `http://localhost:5109/swagger/oauth2-redirect.html`.

PR #591, opened 2018-04-25
Fix swagger oauth redirection error #586

mvelosop:
This PR:
- Stores the new swagger-ui redirection uri to current one as of v3.x for new deployments
- Updates old stored redirection uris to new one.


mvelosop:
This PR closes issues #591 and #562.

Issue #599, opened 2018-05-10
Note on (not so) recent merge conflicts


mvelosop:
### IMPORTANT NOTE ON (NOT SO) RECENT MERGE CONFLICTS

Due to some issues the team of maintainers faced while merging the BFF branch, the dev branch had to be "re-routed", so you might find your repo in a situation like this one:

![image](https://user-images.githubusercontent.com/7357806/39867810-ab396916-544e-11e8-98c5-828ff4bae273.png)

Where:
a) your local **dev** branch is both behind and ahead origin and/or 
b) your previous pending PRs, now have someone else's commit like this:

![image](https://user-images.githubusercontent.com/7357806/39867822-b8f45f5c-544e-11e8-834d-470583158118.png)

In the rest of this issue I'll describe the steps needed to resync your repo and fix the pendind PRs.

**_CAUTION_, there can be many different situations but I am only covering the most general ones, so commit your changes and backup your local repo before attempting any changes.**

### Backgound

On 2018-MAR-29, the **dev** branch had to be reset from commit **0742f941** back to commit **885703c09**, the first one on 2018-MAR-29, and then kept on moving along on 2018-APR-5 with commit **2fb13490**.

So, depending on whether you were working from a fork or from a clone of the main repo and when was the last time you pulled from the repos, you might find in a situation like the one depicted above or something else, kind of "strange".

### Solution - Quick and Dirty

If you just cloned the master eShopOnContainers repo and there's nothing important you want to keep, then just delete your local repo and clone it back again.

Else if you need to sync back you repo, keep reading!

### Solution - Part 1 - Sync repos

So, these SHOULD be the steps to sync back your repo.

The general idea is to:

1. Move your local **dev** branch back to the first commit on 2018-MAR-29 and 
2. Keep moving it forward through the new "road", as depicted here:

![image](https://user-images.githubusercontent.com/7357806/39867846-d0810a9e-544e-11e8-88f7-81028eac1e23.png)

Maybe an analogy helps here, it's just as if you drive into a street and it's blocked, so you have go back and take some other way.

#### IF CLONING FROM THE MASTER REPO

That is, if you cloned the master repo without creaeting a fork (like the image above).

Then, if you cloned the repo on or after 5-APR-2018 you shouldn't need to do anything of what comes next.

If you cloned or forked before 5-APR-2018, keep reading.

##### If cloning the MASTER repo and the top of your local "dev" branch is before 2018-MAR-29

[...]

PR #615, opened 2018-05-23
Master

jimmykani:


mvelosop:
Hi @jimmykani, I guess you created this PR by accident, as there are no comments.

Anyway right now we can't process this PR because of some issues related to #599, that are probably the cause for so many conflicts.

We'll take a look at this after the current version in dev stabilizes and is merged into master first.

Thanks.

Issue #621, opened 2018-05-31
[BUG] Xamarin Applications - Appearing duplicate products in basket after first checkout


joaoGoncalves81:
After adding products to basket and do a first checkout and then continue to add products to basket, the products will appear duplicate.

I guess the problem is in MessagingCenter unsubcribe/subscribe methods in BasketViewModel.

The callback method of the Subscribe in MessagingCenter "AddProduct" is executing more than once in BasketViewModel when user tap on add product button in catalog view. (this only happens after first checkout complete)
This will add more than one product to basket when the user tap on add product on catalog view.

You can try with mock services or real services:

1. Add products to basket
2. Go to basket and checkout
3. Return to catalog view
4. Add one product
5. Go to basket and you will see duplicate items

I tried only on Windows applicaton, but maybe it will happen in Android.
Can you help?

CESARDELATORRE:
@joaofbbg - Thanks for the feedback, we’ll review the issue and respond as soon as possible.
Thanks,

mvelosop:
Hi @joaofbbg, I've been trying to reproduce this and just can't get to this bug, neither on Windows UWP, MVC or SPA.

What is happening, though, is that if you click twice in the same catalog item, it gets added as two separate Order Items, but are summed up on checkout so the order appears as expected: one item, qty: 2.

Could you please double check if this is the situation you are experiencing?

Thanks.

joaofbbg:
Hi @mvelosop , I tried again on Windows UWP after a git pull on dev, and I still have this issue.
Can you please check this video: https://vimeo.com/275411400
You will see after the first order, the items appear in basket are not correct (I'm clicking just once on the catalog) 
Just one side note: on the video you don't see the checkout successful box, but it appear!

joaofbbg:
Hi @mvelosop, I didn't try with Docker, I just Run in Debug Mode, the eShopOnContainers.Windows Project from Visual Studio 2017 and use the Mock Services.

Should I try with Docker? 

New steps to reproduce this errors:

1. Go Cart View and complete a Order.
2. Go to Cart to check if it's empty.
3. Go to catalog and add products
4. Swap between cart and catalog a couple of times
5. Add more products;
6. Check if cart is consistent with the products that you added.

I put video on my google drive now, try this ink: https://drive.google.com/file/d/1ikpQE-37lyOR6k0fI0_sR48UsoOVDYzJ/view

 Thanks

[...]

PR #651, opened 2018-06-28
Fix for issue #621

jacano:
Fix for issue #621
Task.Delay(10) removed from mock services, renderers and View models that could cause potential deadlocks and synchronisation issues, in favour of Task.CompletedTask and Task.FromResult.

### DIFF
--- a/src/Mobile/eShopOnContainers/eShopOnContainers.Core/Services/Basket/BasketMockService.cs
+++ b/src/Mobile/eShopOnContainers/eShopOnContainers.Core/Services/Basket/BasketMockService.cs
@@ -18,52 +18,48 @@ namespace eShopOnContainers.Core.Services.Basket
                 }
         };
 
-        public async Task<CustomerBasket> GetBasketAsync(string guidUser, string token)
+        public Task<CustomerBasket> GetBasketAsync(string guidUser, string token)
         {
-            await Task.Delay(10);
-
             if (string.IsNullOrEmpty(guidUser) || string.IsNullOrEmpty(token))
             {
-                return new CustomerBasket();
+                return Task.FromResult(new CustomerBasket());
             }
 
-            return MockCustomBasket;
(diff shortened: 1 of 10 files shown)


mvelosop:
Hi @jacano,

Thanks for the PR!

I have some general comments:

The first one is to let you know we are following this guidance regarding the use of async/away: http://blog.stephencleary.com/2016/12/eliding-async-await.html, that explicitely discourages the removal of the async/await kewords, for reasons that the author explains in detail in the post.

That guidance was confirmed by the .NET Team (Stephen Toub, Immo Landwerth and Mads Torgersen) and their response was:

“Recommend using async/await by default, and only removing them when doing optimizations because of known perf issues (or you’re implementing libraries you expect to be used on super hot paths). The one exception to this is for public methods that need to do argument validation, in which case just as with enumerables, you generally want to do that validation in non-async methods so that the ArgumentExceptions are propagated out of the method rather than being captured into the returned task.”

So returning Task.FromResult and the like are discouraged, even though you will see them on the codebase (the .NET confirmation is from 20-APR-2018).

The second comment is regarding the issue #621. I must be missing something, because I don't get to find what was causing issue, could you please elaborate?

The final comment is about the removal of several Task.Delay, let's wait for some comment from @davidbritch, who included those lines, just to verify what's their purpose. I'm guessing it's for emulating harsh (or maybe just typical) latency issues with mobile lines, but let's just check and be sure.

davidbritch:
Hi @mvelosop and @jacano 

As you guessed, `Task.Delay` was simulating latency issues. If I remember correctly, originally the delays were substantially longer than in the current code base (more like `Task.Delay(3000)`). I'm happy for them to be removed.

However, I'm also not seeing how this PR fixes the original issue. I've not tried to repo the issue myself yet, but it maybe that there isn't a problem here?

mvelosop:
Thanks for your feedback @davidbritch!

PR #654, opened 2018-07-01
Update mediatr to v5

domenkogler:
According to [Migration Guide 4.x to 5.0](https://github.com/jbogard/MediatR/wiki/Migration-Guide-4.x-to-5.0)

PR #660, opened 2018-07-02
IdentifiedCommand replaced with ExistingCommandBehavior in MediatR pipeline

domenkogler:
IdentifiedCommand replaced with ExistingCommandBehavior in MediatR pipeline.
Implementing response for existing command is now simpler and actual command is passed as parameter.
Added Tests, also for ValidatorBehavior.
#661 Also contains #654


mvelosop:
Hi @domenkogler, this change seems interesting but the domain events processing pipeline had a significant overhaul to handle a set of issues, related to #700, so this code is highly outdated now.

Anyway I'm not sure if it'd be a good idea to add the existing command behavior, because not all domain events are commands that need to have verified for uniqueness.

In the future I'd recommend that before embarking in a PR of this magnitude, you begin a discussion with an issue and check with @CESARDELATORRE, because these changes also have to be reflected in the [Microservices guide/e-book](https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/)

mvelosop:
Closing this PR now but feel free to comment, will reopen if needed.

PR #662, opened 2018-07-03
Convert some magic strings to use "nameof" operator

jo-ninja:
Convert some magic strings to use "nameof" operator


eiximenis:
Merged.
Tanks :)

Issue #699, opened 2018-08-07
Can I send a message with eventbus in foreach loop? 


GloriaGloria:
Hi:
when i try send message in OrderStatusChangedToAwaitingValidationIntegrationEventHandler like this:

   `

      public async Task Handle(OrderStatusChangedToAwaitingValidationIntegrationEvent command)

        {

            foreach (var orderStockItem in command.OrderStockItems)

            {

                var catalogItem = _catalogContext.CatalogItems.Find(orderStockItem.ProductId);

                var hasStock = catalogItem.AvailableStock >= orderStockItem.Units;

                var confirmedOrderStockItem = new ConfirmedOrderStockItem(catalogItem.Id, hasStock);

                var confirmedIntegrationEvent = !confirmedOrderStockItem.HasStock
                ? (IntegrationEvent)new OrderStockRejectedIntegrationEvent(command.OrderId, confirmedOrderStockItem)
                : new OrderStockConfirmedIntegrationEvent(command.OrderId);

                await _catalogIntegrationEventService.SaveEventAndCatalogContextChangesAsync(confirmedIntegrationEvent);
                await _catalogIntegrationEventService.PublishThroughEventBusAsync(confirmedIntegrationEvent);
            }
        }`
everytime  when i send message in this way,  the error occurred like this:

>  the connection is already in a transaction and can not participate in another transaction .
![image](https://user-images.githubusercontent.com/33510889/43748703-9437d76e-9a23-11e8-8ffa-a9abb5546398.png)

Can I write it like that or What should I do in this situation?
Look forward to your reply!
Thanks!

mvelosop:
Hi @GloriaGloria,

It looks like the code you have is not current, it should be like this:

https://github.com/dotnet-architecture/eShopOnContainers/blob/a2bc8aa56777126d34cd0435195e8c040a91aef7/src/Services/Catalog/Catalog.API/IntegrationEvents/EventHandling/OrderStatusChangedToAwaitingValidationIntegrationEventHandler.cs#L29-L43

Hope this helps.

GloriaGloria:
Hi @mvelosop ,
[...]

PR #707, opened 2018-08-14
fix continue send message bugs for issue #699 

GloriaGloria:
when I send message continually or in foreach loop with eventbus , the error occurred like this:
image
Using the same _eventLogService lead to this I believe, so I use a new _eventLogService  when save Event everytime, thus , a new transaction will be created ,so I can send message continually.

I only fix CatalogIntegrationEventService ,and the same problem exists in other IntegrationEventService .

The PR  Assigning closed issue #699 .


mvelosop:
Closing this PR since it's outdated.

The strategy for integration events handling and publishing had a significant revision, so this change no longer applies.

The changes are outlined in the [integration events sample in the wiki logging page](https://github.com/dotnet-architecture/eShopOnContainers/wiki/18.-Centralized-Structured-Logging-in-eShopOnContainers#integration-event-handling).

PR #709, opened 2018-08-15
Changed AspNetCore versions

v-rodudi:
Changed versions: 
Microsoft.AspNetCore.App                                       2.1.0 -> 2.1.3
Microsoft.AspNetCore.Hosting                                 2.1.0 -> 2.1.1
Microsoft.Extensions.Configuration.AzureKeyVault  2.1.0 -> 2.1.1
Microsoft.AspNetCore.TestHost                                2.1.0 -> 2.1.1


mvelosop:
Hi @v-rodudi,

There are some changes here I don't understand, like changing dockerfiles to nightly images, removing environment variables from docker-compose.override, among others.

What's the purpose of this PR?

src/ApiGateways/ApiGw-Base/Program.cs
```diff
@@ -22,7 +22,7 @@ public static IWebHost BuildWebHost(string[] args)
         {
             IWebHostBuilder builder = WebHost.CreateDefaultBuilder(args);
             builder.ConfigureServices(s => s.AddSingleton(builder))
-                .ConfigureAppConfiguration(ic => ic.AddJsonFile(Path.Combine("configuration", "configuration.json")))
+                .ConfigureAppConfiguration(ic => ic.AddJsonFile(Path.Combine("appsettings.json")))
```
v-margko: looks like now we don't need Path.Combine at all

src/ApiGateways/ApiGw-Base/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM microsoft/dotnet:2.1-aspnetcore-runtime AS base
+FROM microsoft/dotnet:2.1.3-aspnetcore-runtime AS base
```
mvelosop: Setting a specific patch version in Dockerfiles would mean that they would have to be updated for every new image release. 

This would probably be a desired production scenario, so versions only change on a schedule and under operations control, but this is not the case in eShopOnContainers, since this is showcase/demo/test app.

Even that "**shouldn't**" be necessary, because we're talking about patch version but then, there are some patches 🙄!

In case the latest stable image (not nightly) causes some problem then the version would be set to the latest known patch that worked fine (as happened at least once), but eventually would be changed back to a fixed minor version.

Anyway let's wait for @CESARDELATORRE's comment on this issue just to confirm.

src/ApiGateways/ApiGw-Base/Program.cs
```diff
@@ -22,7 +22,7 @@ public static IWebHost BuildWebHost(string[] args)
         {
             IWebHostBuilder builder = WebHost.CreateDefaultBuilder(args);
             builder.ConfigureServices(s => s.AddSingleton(builder))
-                .ConfigureAppConfiguration(ic => ic.AddJsonFile(Path.Combine("configuration", "configuration.json")))
+                .ConfigureAppConfiguration(ic => ic.AddJsonFile("appsettings.json"))
```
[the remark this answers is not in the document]
v-rodudi: Thank you! 
When I tried this before, I had an error that the "configuration / configuration.json" file could not be found. 
But now it works. I will revert this change.

mvelosop:
hi @v-rodudi, would you check the CLA, please?
image

src/ApiGateways/ApiGw-Base/Dockerfile
CESARDELATORRE: I'd prefer to use the major version of the image if it already supports AspNetCore 2.1.3

mvelosop: The minor version you mean, right? (i.e. 2.1)

v-rodudi: @CESARDELATORRE, could you please confirm the minor version (2.1) should be used here?

v-rodudi: So, I reverted to 2.1-aspnetcore-runtime in Dockerfiles and checked that it still works for me.

mvelosop:
Closing this PR as it's outdated now.

PR #712, opened 2018-08-18
Made method async

AnriiTokarskyi:
Made call to method GetItemsByIds async to be consistent with async controller action Items.


mvelosop:
Hi @AnriiTokarskyi, thanks for the PR, but it has merge conflicts now, would you solve this so I can merge it?

PR #728, opened 2018-08-29
Change to latest SDK version 2.1.401

VoodooSV:


VoodooSV:
I've had an error message when opened a solution:
"Project file is incomplete. Expected imports are missing"

On my machine was installed latest .NET Core SDK 2.1.401. I've fixed this issue when changed global.json to latest SDK version (2.1.401).

In new projects global.json is absent, maybe it is better to delete this "global.json" at all?

Thanks!

CESARDELATORRE:
Yeah, I just got this issue today in one of the machines that doesn't have installed the .NET Core SDK 2.1.3
I'm merging this PR but we'll probably remove the global.json after confirming with other folks..
Thanks for the heads-up! 👍

unaizorrilla:
Remove global.json is not probably a good idea. This is a good way  to specify the sdk requirement

CESARDELATORRE:
These are the matching rules for the versions in global.json:
https://docs.microsoft.com/en-us/dotnet/core/tools/global-json#matching-rules

@unaizorrilla - The issue with having the global.json for a reference/sample application is precisely the case where folks install a brand new machine with the latest VS and SDKs and we might still have not updated our global.json to the latest SDK version and if the machines where you are trying to run it don't have the older SDK installed (like .NET Core SDK 2.1.3), you get the errors.

Meaning that if using global.json we'll be getting this kind of errors in certain machines (brand new installed from scratch without the older SDKs), in a few months when new SDKs appear...

On the other hand, I agree that this is a restrictive way to specify and control the SDK version that has been tested...

For now, I'll leave the global.json with the updated SDK version (.NET Core SDK 2.1.401) and we'll decide after getting further feedback I'm getting internally with other MSFT teams.
Thanks! 👍

PR #726, opened 2018-08-29
Updated WebSPA to Angular 6, Bootstrap 4, HttpClient and new rxjs, replaced ng2-toastr with ngx-toastr.

IvanFarkas:
Updated WebSPA to Angular 6, Bootstrap 4, HttpClient and new rxjs, replaced ng2-toastr with ngx-toastr.

PR #738, opened 2018-09-02
Set TypeScript ToolSet to version 2.7

rafsanulhasan:
I've specified TS toolset version to 3.0 in the _[WebSPA.csproj](https://github.com/sicsbd/eShopOnContainers/blob/patches/ts3.0/src/Web/WebSPA/WebSPA.csproj)_ file which is a VS issue and has no relations with TypeScript package version from NPM and thus does not conflict with Angular.

### DIFF
--- a/src/Web/WebSPA/WebSPA.csproj
+++ b/src/Web/WebSPA/WebSPA.csproj
@@ -1,102 +1,103 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
-  <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <UserSecretsId>aspnetcorespa-c23d27a4-eb88-4b18-9b77-2a93f3b15119</UserSecretsId>
-    <DockerComposeProjectPath>..\..\..\docker-compose.dcproj</DockerComposeProjectPath>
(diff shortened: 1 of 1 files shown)


raboud:
The SPA actually uses NPM to build and the typescript version should match the one in package.json.
This all should be updated to 2.9.x when the SPA is updated to Angular 6. Version 3 is not currently supported by Angular 6. 

There is another PR for Angular 6 and the SPA. Not sure if they set the typescript version in the csproj.

rafsanulhasan:
@raboud, Open the solution using VS 2017 (15.8.2) and then you'd see that VS expects the WebSPA project to be be updated to use Typescript 3.0 toolset.

You might be right but at you should consider that I've just updated the csproj file to use the TS 3.0 toolset not the **NPM** package nor the **Angular** stuffs.

raboud:
@rafsanulhasan IMHO it should match the NPM version...which will be 2.9 as of PR #726. Plus #726 updates  it to 2.9 in the csproj and packages.json.

rafsanulhasan:
Okay, I've got it. Thanks for your opinion. But I'll still wait for the collaborators'(specially @mvelosop, @CESARDELATORRE and @eiximenis) opinion.

PR #740, opened 2018-09-02
Include/exclude files from git

rafsanulhasan:
exclude following files/folders from github:
1. all **node_modules** folder
2. all wwwroot/lib


mvelosop:
Hi @rafsanulhasan, this is an old timer, let's get going!

Just a question, why did you duplicate the .gitignores in the web apps instead of creating a consolidated .gitignore?

The first option makes the files harder to maintain.

Issue #748, opened 2018-09-04
deploy-ingress and deploy-ingress-azure powershell scripts not working on Mac OS


NileshGule:
I followed the instruction for installing the eShopOnContainers application on Mac OS with CLI as mentioned here
https://github.com/dotnet-architecture/eShopOnContainers/wiki/10.-Setting-the-solution-up-in-AKS-(Azure-Kubernetes-Service)

The Powershell script **deploy-ingress.ps1** fails due to the path separator on Mac.
After changing the directory separator from \ to / it works fine on Powershell on Mac.

```yaml

kubectl apply -f nginx-ingress/namespace.yaml
kubectl apply -f nginx-ingress/default-backend.yaml
kubectl apply -f nginx-ingress/configmap.yaml 
kubectl apply -f nginx-ingress/tcp-services-configmap.yaml
kubectl apply -f nginx-ingress/udp-services-configmap.yaml
kubectl apply -f nginx-ingress/without-rbac.yaml

```

Same issue applies to the **deploy-ingress-azure** Powershelgl script. 
The changes are shown below

```yaml

kubectl patch deployment -n ingress-nginx nginx-ingress-controller --type=json --patch="$(cat nginx-ingress/publish-service-patch.yaml)"
kubectl apply -f nginx-ingress/azure/service.yaml
kubectl apply -f nginx-ingress/patch-service-without-rbac.yaml

```

Is it possible to make the powershell scripts work on Windows and Mac the same way, May be using a `Path.Combine` function?

mvelosop:
Hi @NileshGule, thanks for the PR

However, these files were completely updated so it's outdated, could you redo (and test) this changes on your Mac with the revised files, please?

mvelosop:
Closing issue as stale but feel free to comment, will reopen if needed.

PR #771, opened 2018-09-14
Changed path separator to make powershell script work on Windows and Mac

NileshGule:
Fix for #748 
The powershell scripts is tested on both Mac and Windows 10. With the change in file path separator scripts run fine on both the environments.


turibbio:
I think that this could be a better way to do it: https://github.com/dotnet-architecture/eShopOnContainers/pull/824

mvelosop:
Closing the issue since the code is outdated.

PR #823, opened 2018-10-17
Fix path separator on Mac OS x

turibbio:
On Mac, the original script doesn't work. Thanks to $([IO.Path]::DirectorySeparatorChar), we can easily provide a standard way to execute correctly this script

**Tested on**
Mac OS x High Sierra 10.13.6

**PS Version**                                                                                                                                                                                                                         
PSVersion: 6.0.0-alpha                                                                                                                                                                                               
BuildVersion: 3.0.0.0                                                                                                                                                                                                                                  
GitCommitId: v6.0.0-alpha.9

### DIFF
--- a/k8s/deploy-ingress.ps1
+++ b/k8s/deploy-ingress.ps1
@@ -1,12 +1,12 @@
 kubectl apply -f ingress.yaml
 
 # Deploy nginx-ingress core files
-kubectl apply -f nginx-ingress\namespace.yaml
-kubectl apply -f nginx-ingress\default-backend.yaml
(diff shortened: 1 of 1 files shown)


mvelosop:
Hi @turibbio, I'm closing this PR as it's outdated now, for the code was changed substantially.

Anyway isn't it possible to just use `/` always in the paths instead of `\`?

I've used this "trick" successfully in other contexts but I'm not sure if it works with kubernetes.

turibbio:
no idea, I think my solution is the best you can have without worrying about the platform

PR #827, opened 2018-10-22
58386 Service Fabric Mesh

erikpique:
Service Fabric Mesh: Implement deployment of eShopOnContainers into SF-Mesh.
Add ARM template and scripts for deploy.


eiximenis:
Not apply animore.

PR #855, opened 2018-11-17
Update CatalogItem.cs

sheng-jie:
A small spelling mistake.


mvelosop:
Thanks for fix @ratanparai 👍

PR #860, opened 2018-11-27
add test to CI

erikpique:


benbuckland:
Hey @erikpique, I am very keen to roll this into my project. Interested as to what you did with the compose.yml files, did you just run a separate docker compose command or did you included these to be processed with normal docker-compose files?

PR #876, opened 2018-12-09
Update Order.cs

ceemafour:
Typo in error message when looking up order detail


mvelosop:
Thanks for the fix @ceemafour 👍

PR #902, opened 2019-01-11
Add graphql to project

vahidarr:
I need Add graphql to project to query Api
https://graphql.org/


vahidarr:
I just want pull request.why do you close the request?

mvelosop:
Hi @vahidarr, I closed it because master is a restricted branch, as stated in the [branch guide](https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/branch-guide.md)

vahidarr:
So can i ask it in Issues???

mvelosop:
Well, it's not really a matter of just asking for it, since this is sample app for microservices architecture and patterns that goes along with the microservices guide ([online](https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/) and in [pdf](https://aka.ms/microservicesebook)).

I think GraphQL might be an interesting option, similar to an [API Gateway/Aggregator](https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/architect-microservice-container-applications/direct-client-to-microservice-communication-versus-the-api-gateway-pattern), but it's not REST and AFAIK it's not currently in the backlog.

I'm not even sure if GraphQL makes sense, from an architectural point of view (see [this excellent comparison](https://blog.goodapi.co/rest-vs-graphql-a-critical-review-5f77392658e7)).

Perhaps @CESARDELATORRE might comment on this.

CESARDELATORRE:
I'm not very keen on GraphQL.. That comparison provided by Miguel gives quite a few reasons similar to the ones why we didn't add it to eShopOnContainers.

It is also not mainstream when using ASP.NET Core Web API.

However, I'm still open to discussions. If anyone provides good reasons (more pros than cons) we could take it into account for a future backlog. Bit I have my doubts about it. Still, open to discuss. 👍

PR #934, opened 2019-02-06
OrderController Recommendation

mohsenalikhani:
It's good idea to take advantage of GetOrderByIdQuery and handle by GetCustomerByIdQueryHandler


mvelosop:
Hi @mohsenalikhani, thanks for the suggestion!

I think it'd certainly be a interesting example to handle Queries in the MediatR pipeline.

There's not an elegant way to handle this in MediatR right now, but it's in the works, don't know how long it'll take though (you can take a look at https://github.com/jbogard/MediatR/issues/284 and https://github.com/aspnet/Extensions/pull/536)

Anyway I'll merge this right away!.

PR #966, opened 2019-03-10
host don't need anymore

SychevIgor:
hosts file dropped


mvelosop:
Thanks for the PR @SychevIgor, LGTM, merging now!

Issue #753, opened 2018-09-06
Is there any plan to use the integrated event library?


yang-xiaodong:
Hello, eShop team:

I created a library CAP ( [https://github.com/dotnetcore/CAP](https://github.com/dotnetcore/CAP) ) about the integration event of EventBus, which has been created for more than one year, currently there are 13k+ downloads in NuGet. It has been running in our production environment for a few months and works well.

There are some bugs (#721, #700) in the eShop integrated event module. In CAP, we solve these problems and implements the Outbox pattern described in the [eShop ebook](https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/multi-container-microservice-net-applications/subscribe-events#designing-atomicity-and-resiliency-when-publishing-to-the-event-bus) with  atomicity and resiliency.

CAP currently supports Kafka and RabbitMQ as the message queues, and support SQL Server, MySQL, PostgreSQL, and MongoDB 4.0+ as microservices business database.

As a Microsoft MVP, I hope more people can use the library and be able to help them, If eShop is willing to use it, it will spread faster. :)

If you are interested, I would be happy to submit a PR to help eShop complete the integration work.

Please let me know if you have any questions, looking forward to your reply.

eiximenis:
Adding @CESARDELATORRE

yang-xiaodong:
no response , close

yang-xiaodong:
Reopen for discussion

mvelosop:
Hi @yang-xiaodong, please see [my comment on PR #970](https://github.com/dotnet-architecture/eShopOnContainers/pull/970#issuecomment-472833963)

mvelosop:
Closing this issue now, since this will be handled as per @CESARDELATORRE's [comment](https://github.com/dotnet-architecture/eShopOnContainers/pull/970#issuecomment-474924473)

Issue #700, opened 2018-08-08
[BUG] There seems to be an atomicity issue while rising domain events in Ordering.Api


arielmoraes:
In the Ordering service we have a `ValidateOrAddBuyerAggregateWhenOrderStartedDomainEventHandler` which handles the `OrderStartedDomainEvent`, after the Buyer is added or updated a call to `SaveEntitiesAsync ` is made to dispatch the domain events and save the changes to the database, after that, a call to `PublishThroughEventBusAsync` is made. So, if the changes was already saved by calling the `SaveEntitiesAsync ` method no atomicity will be in place or am I missing something? There is even a bigger problem if a chain of domain events are raised because only the last call to `PublishThroughEventBusAsync` will have atomicity between the current handled aggregate and the event.

mvelosop:
Hi @arielmoraes, that's a very good question! 👍

And the fact is that you're right, there's no global atomicity between microservices and that's actually on purpose!

Doing so would require the use of distributed transactions, using two phase commits, but that would cause coupling between the microservices, which is one of the key issues that we want to avoid in a microservices architecture, where each microservice has to be as independent as possible.

So, in a microservices architecture we are actually looking for eventual consistency by using an asynchronous message-based communication system, to propagate changes between microservices, in a pub/sub pattern.

And, just as you realized, doing so in a resilient way is very hard and requires more complexity than the one handled in eShopOnContainers right now.

Some ways to achieve the required resiliency for a production application would be using:

1. The [OutBox pattern](http://gistlabs.com/2014/05/the-outbox/)
2. [Event sourcing](https://docs.microsoft.com/en-us/azure/architecture/patterns/event-sourcing) or
3. [Transaction log minning](https://www.scoop.it/t/sql-server-transaction-log-mining)

eShopOnContainers is just a sample to showcase architectural patterns and is not intended to be used in production as is, so the implementation of the Pub/Sub pattern is a basic one based on RabbitMQ.

You can get more details about this issue in @CESARDELATORRE's [Microservices Architecture eBook](https://aka.ms/microservicesebook) on sections:

* Asynchronous message-based communication (p. 57) and
* Implementing event-based communication between microservices (integration events) (p.131)

Hope this helps.

arielmoraes:
[...]

PR #970, opened 2019-03-14
[Features] Introducing a new EventBus CAP for eshop

yang-xiaodong:
Hello  :

I am the creator of [CAP]( https://github.com/dotnetcore/CAP), and also a Microsoft MVP who loves open source very much. I saw eShop's EventBus a few months ago. It looks a bit simplistic and I found a lot of EventBus related in the issue list, and I answered some of them.

Last year, I opened an issue(#753) to ask if there was a plan to use a third-party event integration library, but I was frustrated that I didn't get more responses.

Recently, I had more time to do the extra works. I spent some time completing the CAP integration work for eshop and had already done the tests. He worked very well and saved a lot of code.

Here are the main changes I made:

- Removed the EventBus building block
- Replaced all project EventBus to CAP
- Keep the original code structure as much as possible

Let me know if PR needs to be merged into a new feature branch.

I will reopen the issue(#753) so that we can discuss.

Looking forward to your feedback.   

 @mvelosop @CESARDELATORRE @eiximenis


alexinea:
It is pleasure to hear that CAP has been integrated into the sample project eShopOnContainers. 😁

mvelosop:
Hi @yang-xiaodong, Thanks for dedicating so much effort to get into the eShopOnContainers' internals.

There are several issues with this PR that deserve some special comments.

- This is definitely a LARGE and impactful PR, not only by its sheer size (almost 140 files), but because it touches most, if not all, important microservices. It also affects the Microservices Guide/e-book.

- Something that calls my attention is that you replaced the existing EventBus abstractions with yours. And that's the main cause for the size of the PR. 

  I know I'm probably missing something important, but what's the reason for such a change? I mean, why not start with a concrete implementation using CAP? (BTW, I'm not suggesting you to do it).

- Transaction atomicity and chained events handling (issue #700 and related) had a significant overhaul recently, and one of the important changes was that the resulting integration events are only published if the domain events chain transaction completes successfully.

  It looks like this feature was removed, but then again, I just might have missed it, for I didn't get into all details. I focused mainly on Ordering.API.

- There are also some seemingly minor changes to logging contexts, for no apparent reason, that actually break integration events correlation and traceability.

- As a final comment, it looks like CAP hides away several details and completely replaces RabbitMQ and Azure ServiceBus implementations (you actually deleted them in this PR), which might be fine in another context, but I'm not sure this is a good thing given the purpose of eShopOnContainers.

Anyway, this just my (biased) opinion and the final decision is, of course, on @CESARDELATORRE .

Cheers.

cc/ @nishanil

yang-xiaodong:
Hello @mvelosop ,
Thanks for your reply, I am very happy to discuss some of your questions.

---
- > This is definitely a LARGE and impactful PR, not only by its sheer size (almost 140 files), but because it touches most, if not all, important microservices. It also affects the Microservices Guide/e-book.

Yes , this is really a very large PR, and I understand that the large PR is difficult to accept for the open source projects, but I think it is worth it. 
I have seen most of the content in e-books, especially the outbox pattern description of EventBus and CAP is implemented this pattern.

---

- > Something that calls my attention is that you replaced the existing EventBus abstractions with yours. And that's the main cause for the size of the PR.

Yes, I removed the EventBus building blocks and replaced it with the abstraction provided by CAP. It is roughly divided into two steps :
1. The event handler adds the `[CapSubscribe]` attribute to subscribe to specific events and inherits the `ICapSubscribe` interface so that the CAP can find it faster when it starts.
2. Add the event handler class to the DI.

---

- >Transaction atomicity and chained events handling (issue #700 and related) had a significant overhaul recently, and one of the important changes was that the resulting integration events are only published if the domain events chain transaction completes successfully.

When the CAP startup, it generates the inbox table and the outbox table to the microservices database to ensure atomicity with the local transaction operations, and the CAP will take over the transaction object of the database context in the business operation to guarantee and store the atomicity of the event operation.

You can see here how to implement atomic operations between business code and storage events at [here](https://github.com/dotnetcore/CAP#publish).

Of course, the integration events are only published if the domain events chain transaction completes successfully. Their implementation is slightly different for different databases. For example, in SQL Server, we use the Diagnostics feature provided by the SQL Client to detect the commit operation of the transaction and the code is [here](https://github.com/dotnetcore/CAP/blob/master/src/DotNetCore.CAP.SqlServer/Diagnostics/DiagnosticObserver.cs#L47).

---

- >There are also some seemingly minor changes to logging contexts, for no apparent reason, that actually break integration events correlation and traceability.

Inside the CAP, there is a wrapper around the integration event  transport object. We can track the event associations in the database based on the internal number, and CAP provides a Dashboard UI for user to search.

---

- > As a final comment, it looks like CAP hides away several details and completely replaces RabbitMQ and Azure ServiceBus implementations (you actually deleted them in this PR), which might be fine in another context, but I'm not sure this is a good thing given the purpose of eShopOnContainers.

Yes, CAP hides some details about the implementation of EventBus, and provide different message queues and different event storage extension packages.

---
In this PR, there are some integration events that are used the in-memory storage extension package., because these microservices use databases such as Redis that do not support ACID, obviously they can't guarantee atomicity.

MongoDB supports ACID transaction operations since version 4.0+, and CAP has provided support for it, but the MongoDB currently used by eshop is not a cluster, so it can not used and I use in-memory storage instead.

paulvanbladel:
@yang-xiaodong @mvelosop 
I didn't know CAP (well,... I know the CAP theorem) until I saw this PR.
I did today some monkey testing with a small sample with rabbitmq. So, I'm just sharing with you my findings.
- configure cap with 2 apps pointing to same sql database for the outbox tables but in different schema's.
- a simple event class with event handler
- applied different disconnect scenarios, also when rabbitmq is offline.

CAP worked outstanding !
The current integration event handling solution in the eShopOnContainers is good for learning purposes but very imcomplete (the current version has serious problem with disconnect scenarios). CAP would definitely bring things to yet another level.

CESARDELATORRE:
Adding @eiximenis and @unaizorrilla so they can also evaluate and provide feedback.

I think it might not make sense to put CAP under our EventBus abstractions since CAP already has its opinionated interfaces offering further features and also works on top of Azure Service Bus, RabbitMQ and Kafka. It wouldn't make sense to have two layers of abstractions, right?

However, eShopOnContainers aims to provide patterns implementation for learning purposes. Being prescriptive with CAP or any other Service Bus is a different thing..  

Why **CAP** and not **NServiceBus**, **MassTranssit** or any other Service Bus approach?

@yang-xiaodong - What you could do is to fork eShopOnContainers and have that implementation in a different repo, such as the **NServiceBus eShopOnContainers** implementation here:

https://github.com/Particular/eShopOnContainers 

So we could also point to your CAP implementation from the documentation and the GitHub repo.
That way you can maintain it up-to-date related to CAP future versions, etc.

Thoughts?

yang-xiaodong:
@CESARDELATORRE Thanks for your reply, I agree with you about that current patterns implementation for learning purpose.

 I will fork the eShopOnContainers and provide EventBus integration with CAP. In fact, it is the [repository](https://github.com/yang-xiaodong/eShopOnContainers) pointed to by the current PR and I have changed the default branch to CAP.

Should I submit a PR that modifies README to point to my repository?  Can accept？

CESARDELATORRE:
Sure, let’s do that PR for the  README.
Thanks! :)
Cesar

________________________________

yang-xiaodong:
Hello @CESARDELATORRE, 

I have submitted a PR (#975) about adding EventBus note to the README.  I will complete the description in my eShopOnContainers branch later.

Thanks ! :)

mvelosop:
Hi @yang-xiaodong, I'm closing this PR now, just waiting for your update to PR #975

Issue #612, opened 2018-05-22
[Bug?] EventBusRabbitMQ always creates ConsumerChannel even there is no subscription exists


cilerler:
As you may see it in here https://github.com/dotnet-architecture/eShopOnContainers/blob/e3daacb9f63adb361776c3fc888612d5fa4ae06a/src/BuildingBlocks/EventBus/EventBusRabbitMQ/EventBusRabbitMQ.cs#L41 **ConsumerChannel** gets created regardless of if there is a subscription exists or not.

This behavior led application to retrieve the message right after publish. Would you please confirm that if this is an intended behavior please?

fbarajas-fmg:
I am running into the same issue. Is this intended? I feel like it might be a bug.

zoinkydoink:
I have the same issue, which is causing my actual subscriber to only pick every second created event, one  as stated here

https://github.com/dotnet-architecture/eShopOnContainers/issues/911

mvelosop:
Hi @cilerler,

Right now I'm adding a centralized logging sink to eShopOnContainers and perhaps this could be a good issue to try it out.

But I don't understand the scenario where this problem occurs in eShopOnContainers and what's the simptom.

Could you describe the scenario so we can have a test case? 

I'd like to identify where we should set logging traces to find out where the problem is.

Thanks.

cilerler:
Hi @mvelosop 

Consumer should only connect if there is a subscriber and if there is no handler for that message type, it should reject the message _(and if there is DLX/DLK on that queue it will end up on another queue but it is a different topic)_

As you may see here https://github.com/dotnet-architecture/eShopOnContainers/blob/e3daacb9f63adb361776c3fc888612d5fa4ae06a/src/BuildingBlocks/EventBus/EventBusRabbitMQ/EventBusRabbitMQ.cs#L184-L203

it calls the `ProcessEvent` to process it and even if there is no subscription _(based on code below)_ it sends an acknowledge.

https://github.com/dotnet-architecture/eShopOnContainers/blob/e3daacb9f63adb361776c3fc888612d5fa4ae06a/src/BuildingBlocks/EventBus/EventBusRabbitMQ/EventBusRabbitMQ.cs#L208-L210

Meaning that message has never been processed and it is gone forever.

Please let me know if you need further details.

_(Thank you for picking this one as a test case)_

mvelosop:
Thanks @cilerler, 

[...]

Issue #500, opened 2018-02-03
[BUG] Losing messages in Azure Service Bus subscription


i02coroj:
Hi.
I've realized there is a case when messages published into Azure Service Bus can be never handled.
This is the scenario:
- The service (for instance Basket) is down.
- Some events of different types are published into the service bs.
- Basket service gets started.
- When creating the subscription, there is a condition to only process messages if there is already a subscription saved in memory. But the message is still set as completed
- This means that other type events are directly marked as completed even when they haven't been processed.

It can be solved just checking out the message has been actually processed and mark it as completed only in that case.
Please, let me know if you want me to make a pull request with it or it is something you already knew and you want to let as it is.

CESARDELATORRE:
Yes please, submit a PR with the fix and explain the issue in the PR and the functional tests you did after fixing it, ok?
Thanks a lot! 👍

i02coroj:
Hola César.
Sorry for my late answer on this.

The issue happens when you are subscribed to several kind of events.
For instance, in Basket.API you have two of them:
            eventBus.Subscribe<ProductPriceChangedIntegrationEvent, ProductPriceChangedIntegrationEventHandler>();
            eventBus.Subscribe<OrderStartedIntegrationEvent, OrderStartedIntegrationEventHandler>();

If an event is received when the service is up everything is ok.
But if the service is down and some messages are published, when the service starts up they can be lost.

That happens because inside Subscribe, pending events are processed but only if there is already a subscription for it.
So, in this example, when subscribing to ProductPriceChangedIntegrationEvent, if there is any OrderStartedIntegrationEventHandler it will be lost, as there is no subscription to it yet, but it is marked as completed anyway.

To solve it, I've added a single line where the event is marked as completed only if actually completed.
I've tried to make a pull request but I get an error 403 when pushing my changes to a fix branch. So, please let me know how I should process.

raboud:
Are you trying to push to branch of the main repo?

Most of us do not have write privileges to the main repository. 

I fork of the repository, I then make a branch in the forked repository and push changes to the branch in the forked repository. I then do the PR from the branch in the forked repository.

i02coroj:
Hi. Sorry for my late answer.
Thanks for that, I just did it.

Cheers!

paulvanbladel:
I'm experiencing similar problem in the rabbitMq version.
Is the problem not that during ProcessEvent the Subscriptionmanager is already used before it is actual initialized?
[...]

PR #972, opened 2019-03-15
Split RabbitMQ channel create and consumer create.

liubaishui:
Because when channel create the Subscribe is not register, so the previously stored message may lose. This fix split the channel create and consumer create, when Subscribe registered then call the consumer create function.

### DIFF
--- a/src/BuildingBlocks/EventBus/EventBusRabbitMQ/EventBusRabbitMQ.cs
+++ b/src/BuildingBlocks/EventBus/EventBusRabbitMQ/EventBusRabbitMQ.cs
@@ -98,7 +98,7 @@ namespace Microsoft.eShopOnContainers.BuildingBlocks.EventBusRabbitMQ
 
                     channel.BasicPublish(exchange: BROKER_NAME,
                                      routingKey: eventName,
-                                     mandatory:true,
+                                     mandatory: true,
                                      basicProperties: properties,
                                      body: body);
                 });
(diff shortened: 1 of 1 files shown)


liubaishui:
Relate with #612 and #500

mvelosop:
Thanks @liubaishui, this seems to be working great! 👍

Just in case someone wants to check this, I could reproduce the error with these steps:

1. Start app with \
  `docker-compose up`
2. Buy an item and go to checkout
3. Place an order
4. Stop the catalog.api container with \
  `docker container stop $(docker ps -qf ancestor=eshop/catalog.api)`
5. Open the RabbitMQ manager at: http://10.0.75.1:15672/#/queues/%2F/Catalog until a message is stuck in the Catalog queue, you should see something like this:
  image
6. Shutdown the whole app with [Ctrl+C]
7. Startup Seq with \
  `docker-compose up seq`
8. Startup RabbitMQ with \
  `docker-compose up rabbitmq`
9. Wait for Seq and Rabbit to spin up
10. Open RabbitMQ at: http://10.0.75.1:15672/#/queues/%2F/Catalog, you should see an image similar to the last one, with the message still stuck in the Catalog queue.
11. Start catalog.api with \
  `docker-compose up catalog.api`
12. Look at RabbitMQ until the message is delivered (the queue drops to zero)
13. You should something like this in Seq:
  image
  You can see that the message was not handled by catalog.api and disappeared from the queue, so it was lost.

If you repeat the above procedure after this PR, you should get some log traces like the following:
  image
  Where you can see that the pending message was handled upon starting of catalog.api, so it wasn't lost.

Thanks @liubaishui 😊

PR #983, opened 2019-04-01
Fix markdown headers

ratanparai:


mvelosop:
Thanks @ratanparai!

Will merge as soon as validations complete!

PR #984, opened 2019-04-02
Fix typo

### LINKED ISSUE #982: Typo on the order detail page
Minor detail but maybe something to look at for next update :)

It says on the order detail page:
'The payment was performed at a simulated "American Bank checking bank account **endinf** on XX35071"'

Where I expected
'The payment was performed at a simulated "American Bank checking bank account **ending** on XX35071"'

rafsanulhasan:
also fixes #982

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
 
+# Project Launch Settings
+Properties/launchSettings.json
+
 # Build results
 [Dd]ebug/
 [Dd]ebugPublic/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -11,6 +11,9 @@
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
 
+# Project Launch Settings
```
mvelosop: Why excluding launchSettings from source control?

rafsanulhasan: It is because because whenever I reload the solution, all the launch settings automatically marked as changed to but diff shows there is absolutely nothing changed.

However you can ignore this and include those in the source control you that is needed.

rafsanulhasan:
Hi @mvelosop I reverted the .gitignore file back to its original state. is it okay now?

mvelosop:
Thanks @mvelosop, merging now!

Issue #848, opened 2018-11-12
Library Manager manifest contains syntax errors.


zeekay18:
I'm getting this error when building the project the first time.

Severity	Code	Description	Project	File	Line	Suppression State
Error	LIB004	Library Manager manifest contains syntax errors. Please fix the errors in libman.json, then try again.	Identity.API	C:\Users\Azeez\source\eShopOnContainers\src\Services\Identity\Identity.API\libman.json	1

chanaka1:
This is a Issue of Visual studio. Check the [Library Manager Reported Bug](https://github.com/aspnet/LibraryManager/issues/111)

mvelosop:
Hi @zeekay18 are you still getting this issue?

I have no problem with VS 15.9.6 in any of the projects that use libman.

mvelosop:
Closing this issue as it seems stale, but feel free to comment, will reopen if needed.

rafsanulhasan:
I'm still facing this issue in VS 15.9.10

rafsanulhasan:
I found two solutions:
1. Removing the Nugget package: Microsoft.Web.LibraryManager.Build
2. Removing the **files** array from libman.json

I think the 2nd one is good enough.
How about submitting a PR on this @mvelosop ?

PR #986, opened 2019-04-02
Fix warnings

### LINKED ISSUE #985: Some warning and issues
Hello, 
I'm using using VS 2017 15.9.10 and getting some warnings and errors.
1. Getting errors while building the project, why do we need to run build task for library manager since all the required files are already on the repo (#848)
2. Compiler gives warning about **Razor syntax** for rendering partial pages to be changed to Tag Helpers
3. The NuGet package version conflict: Microsoft.ApplicationInsights
4. Using version in Microsoft.AspNetCore.App
5. Called some deprecated methods in authorize filters & obsolete methods in logging Adding ApplicationInsights in Startup
6. There are some async/await warnings

rafsanulhasan:
fix #985


src/Services/Identity/Identity.API/Identity.API.csproj
```diff
@@ -28,7 +28,6 @@
     <PackageReference Include="Microsoft.AspNetCore.HealthChecks" Version="1.0.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.AzureKeyVault" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.AzureAppServices" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Web.LibraryManager.Build" Version="1.0.172" />
```
mvelosop: Can't remove building the client-side libraries, because they are not included in version control (as usual) so there's no guarantee that they'll be there when building the project.

For example, you have them in your computer because you built the project some time ago, but this would work if yu fresh cloned the project and tried building the project with Docker.

You'd have to build it first in VS, and that's nonsense when working with Docker, so that's why the building from LibMan is important.

src/Web/WebMVC/WebMVC.csproj
```diff
@@ -31,10 +31,9 @@
     <PackageReference Include="Microsoft.AspNetCore.HealthChecks" Version="1.0.0" />
     <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.AzureAppServices" Version="2.2.0" />
-    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.2.3" />
     <PackageReference Include="Microsoft.AspNetCore.DataProtection.Redis" Version="2.2.0-preview2-35157" />
     <PackageReference Include="Microsoft.VisualStudio.Azure.Fabric.MSBuild" Version="1.6.7" />
-    <PackageReference Include="Microsoft.Web.LibraryManager.Build" Version="1.0.172" />
```
mvelosop: Restore LibMan Builder

src/Web/WebStatus/WebStatus.csproj
```diff
@@ -22,7 +22,6 @@
     <PackageReference Include="Microsoft.AspNetCore.Diagnostics.HealthChecks" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.AzureKeyVault" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.AzureAppServices" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Web.LibraryManager.Build" Version="1.0.172" />
```
mvelosop: Restore LibMan Builder

src/Web/WebMVC/WebMVC.csproj
rafsanulhasan: Ok, if it makes sense. However it that case, I think the files array that specifies _the files to be downloaded_ should be removed because this is the second solution until it is fixed

mvelosop: I'm sorry, I'm afraid I don't understand.

Why it's necessary to remove the files array?, those are just the files needed and, what has to be fixed regarding this?

rafsanulhasan: because during build, the manifest file of libman, shows an error (see #848), which only fixes in this two ways: [see details here](https://github.com/aspnet/LibraryManager/issues/111)

rafsanulhasan: @mvelosop, and this can only be resolved by either of the following:
1. Removing LibraryManagerBuild NuGet package (which is not right as you suggested)
2. removing the files array which specifies the files to be downloaded by LibMan
3. manual workflow
   a. putting the files into **.node_nodules** folder, using _filesystem_ as a source, specifying those files from

ratanparai:
This `PR` is getting massive. I think you should create separate `PR` for multiple things that are not related with one another if @mvelosop agree with me.

PR #989, opened 2019-04-04
add editor config

rafsanulhasan:


mvelosop:
Hi @rafsanulhasan, thanks for the PR.

I think adding editor config is a good idea, specially thinking on VS 2019, but please, add only that in the PR, don't understand the change in GUIDs in some projects and that means unnecessary testing.

Cheers.

rafsanulhasan:
Hi @mvelosop, fixed it. 

Just to confirm you, I didn't make any change to project GUIDs manually. However, thank you for letting me know about this. This happened only because of absence of mind. I should have checked the diff before committing. I will take care of it from the next time.

mvelosop:
@rafsanulhasan, this PR has changes other that related to .editorconfig!

PR #994, opened 2019-04-04
fixed libman errors

### LINKED ISSUE #985: Some warning and issues

rafsanulhasan:
I think the files array that specifies _the files to be downloaded_ should be removed because this is the second solution until it is fixed

because during build, the manifest file of libman, shows an error (see #848), which only fixes in this two ways: [see details here](https://github.com/aspnet/LibraryManager/issues/111)

Note that I'm using VS 2017 15.9.11. However in VS 2019 16.0.0, this issue is resolved, but I think this should not be IDE dependent

and this can only be resolved by any of the following:
1. Removing LibraryManagerBuild NuGet package (which is not right as you suggested)
    So I restored the package.
2. removing the files array which specifies the files to be downloaded by LibMan
3. manual workflow (it would be redundant)
   a. putting the files into **.node_nodules** folder (either manually or by node packages as discussed in [this issue](https://github.com/aspnet/LibraryManager/issues/111)) , 
   2. using _filesystem_ as a source, 
   3. specifying those files from **.node_modules** directory for that library

fixes #985

### DIFF
--- a/src/Services/Identity/Identity.API/libman.json
+++ b/src/Services/Identity/Identity.API/libman.json
@@ -9,14 +9,6 @@
     {
       "provider": "unpkg",
       "library": "bootstrap@4.1.3",
-      "files": [
-        "dist/css/bootstrap.css",
-        "dist/css/bootstrap.css.map",
-        "dist/css/bootstrap.min.css",
-        "dist/css/bootstrap.min.css.map",
-        "dist/js/bootstrap.js",
-        "dist/js/bootstrap.min.js"
-      ],
       "destination": "wwwroot/lib/bootstrap/"
     },
     {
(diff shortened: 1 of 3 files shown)


mvelosop:
Hi @rafsanulhasan,

Thanks for separating the PR in issues and to @ratanparai for the suggestion.

I just can't reproduce the error you mention.

My setup:

- VS: 15.9.11
- SDK: 2.2.202
- Runtime: 2.2.3
- Commit: 6396e0ffbe53555d1d1b03bf6a1bac2bd6d80cac

I just get:

image

or:

image

rafsanulhasan:
@mvelosop I have the same setup as yours.

mvelosop:
I have just the same extension, perhaps some leftover from a previous compilation.

Closing this issue then.

PR #997, opened 2019-04-05
Logout post action from _LoginPartial.cshtml

LeeDumond:
Fixed broken reference to non-existent Logoff action. Changed to post to Logout action.


mvelosop:
Closing now will be redone as commented in #999

PR #998, opened 2019-04-07
fix missing reference that was causing Catalog Functional Tests to fail -- see https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2#test-app-prerequisites

LeeDumond:
fix missing reference that was causing Catalog Functional Tests to fail -- see https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2#test-app-prerequisites


mvelosop:
Closing now will be redone as commented in #999

PR #1000, opened 2019-04-09
fix references in functional test projects

LeeDumond:
fix missing references that were causing functional tests to fail.


mvelosop:
Thanks @LeeDumond, tests for this projects run fine now.

Merging now!

PR #1001, opened 2019-04-09
change form action from non-existent Logoff to Logout

LeeDumond:
Logout form will now point to the correct action.


mvelosop:
@LeeDumond I doesn't look like this really is used anywhere, but anyway you're right, about the correct action.

Thanks.

PR #1004, opened 2019-04-12
Get_get_campaign_by_id now fails initially

LeeDumond:
This test is looking for ID = 81, but the test fails when the DB is first built, because the original seed only populates 2 campaign rows. By changing this to ID = 2, the test will pass without having the other POST/PUT tests run several times first.

### DIFF
--- a/src/Services/Marketing/Marketing.FunctionalTests/CampaignScenarios.cs
+++ b/src/Services/Marketing/Marketing.FunctionalTests/CampaignScenarios.cs
@@ -27,7 +27,7 @@ namespace Marketing.FunctionalTests
         [Fact]
         public async Task Get_get_campaign_by_id_and_response_ok_status_code()
         {
-            var campaignId = 81;
+            var campaignId = 2;
             using (var server = CreateServer())
             {
                 var response = await server.CreateClient()


mvelosop:
Hi @LeeDumond, yep I saw that one, but could not investigate further, because there's another test that adds an item with Id = 81

That's not a good practice, but I haven't checked in detail for the reason for it to be so, did you catch it?

LeeDumond:
Yes, these are functional tests, so they run against the REAL database, not
a mocked data repository. Hence, the test depends on real data being
present. In this case, the test will fail if there is no campaign with Id =
81.

When the database is first built, it seeds only two rows, with Id = 1 and
Id = 2. So, since there is no row with ID = 81, the test fails.

If you run ALL the campaign functional tests, there are POST, PUT, and
DELETE tests in there as well. So, after the first test run, there will be
the following Ids: 1,2,21,31

After the second test run: 1.2.21.31.51.61

After the third: 1,2,21,31,51,61,81,91

So then the test in question will now start passing (because there is a row
with ID = 81).

Obviously this only would be true if all the tests are run together. If any
of the POST, PUT, or DELETE tests are run individually, it would throw off
the expected sequence of Ids.

So obviously this is a brittle test, but functional tests are often quite
brittle since they run against real data that you cannot assure will be
there. But at least my PR is an improvement since it does not rely on the
tests being run several times before they all start to pass.

The test could be made more robust by first checking if there is any data,
and then setting the Id to the first ID found, or asserting the test to be
inconclusive if there are no data at all. If I were writing this test for
my own project, that is what I would do.

-- Lee

mvelosop:
I agree with you, so I'll merge this and notify @erikpique, just in case we are missing something important.

Thanks 😊

PR #1008, opened 2019-04-17
Fix marketing functional tests that rely on existing campaign

LeeDumond:
There are 2 tests that use campaignId = 81 that fail when a campaign with that Id does not exist. This is exactly the same scenario as seen in PR #1004.


mvelosop:
Thanks @LeeDumond,

LGTM so I'm merging it now.

Cheers 😊

PR #1009, opened 2019-04-18
model declarations in razor views resulting in unsolved references

LeeDumond:
In some razor views, the model declaration is embedded inside a code block, which makes it difficult for Visual Studio to resolve the model references. This PR moves the model declaration for the affected views outside of the code blocks.

### DIFF
--- a/src/Web/WebMVC/Views/Campaigns/Details.cshtml
+++ b/src/Web/WebMVC/Views/Campaigns/Details.cshtml
@@ -1,6 +1,8 @@
+@model CampaignItem
+
 @{
     ViewData["Title"] = "Campaign details";
-    @model CampaignItem
+
     var headerList= new List<Header>() {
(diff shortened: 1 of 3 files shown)


mvelosop:
Hi @LeeDumond, thanks for the heads up!

This doesn't seem to cause any problem, at least not to me and haven't seen an issue either.

Anyway, it's certainly not a common place for the model, so I think your change is ok and I'll merge it now.

Cheers 😊

LeeDumond:
I don't think it actually causes a problem with compilation but it makes ReSharper complain at least :)

mvelosop:
Oh! that can be considered a valid reason certainly!

Although I opted to uninstall R# (response time was a issue too) and give [Roslynator](https://github.com/JosefPihrt/Roslynator) a try, you might like it. I haven't missed R# since.

Cheers.

PR #1010, opened 2019-04-18
add missing Error view for WebStatus project

LeeDumond:
The Error action in the Home controller in the WebStatus project returns a View that does not exist. This PR adds a simple Error view based on the default error view from the .NET Core MVC template.

### DIFF
--- a/src/Web/WebStatus/Views/Shared/Error.cshtml
+++ b/src/Web/WebStatus/Views/Shared/Error.cshtml
@@ -9,18 +9,15 @@
 <head>
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Error</title>
+    <link rel="stylesheet" href="~/lib/bootstrap/dist/css/bootstrap.min.css" />
 
     <environment names="Development">
-        <link rel="stylesheet" href="~/lib/bootstrap/dist/css/bootstrap.css" />
         <link rel="stylesheet" href="~/css/site.min.css" type="text/css" />
-
     </environment>
     <environment names="Staging,Production">
-        <link rel="stylesheet" href="https://ajax.aspnetcdn.com/ajax/bootstrap/3.3.7/css/bootstrap.min.css"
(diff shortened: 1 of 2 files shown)


src/Web/WebStatus/Views/Shared/Error.cshtml
```diff
@@ -0,0 +1,40 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Error</title>
+
+    <environment names="Development">
+        <link rel="stylesheet" href="~/lib/bootstrap/dist/css/bootstrap.css" />
+        <link rel="stylesheet" href="~/css/site.min.css" type="text/css" />
+
+    </environment>
+    <environment names="Staging,Production">
+        <link rel="stylesheet" href="https://ajax.aspnetcdn.com/ajax/bootstrap/3.3.7/css/bootstrap.min.css"
```
mvelosop: Since this app is currently using bootstrap 4.1.3 I think it'd be better to either reference the local bootstrap.min.css from wwwroot or the 4.1.3 from the cdn.

In this case I think the wwwroot version would be just enough.

LeeDumond:
Excellent point, will take care if that asap

src/Web/WebStatus/Views/Shared/Error.cshtml
LeeDumond: Ah, it appears the there is no such version being hosted on the Microsoft CDN. Theirs jumps from 4.1.1 to 4.2.1. Should we use a different CDN? The _Layout.cshtml files in other projects use 3.3.7, which is what I used as a guide. Maybe use 4.1.1 in this case?

LeeDumond: Or I will take your suggestion and just use the local...

mvelosop:
LGTM @LeeDumond, I'm merging now.

Thanks!

Issue #295, opened 2017-08-26
can not show login page  on mvc site 


infofromca:
just click log in ,  go to http://localhost:5100/Account/SignIn, then get this 
![login-err](https://user-images.githubusercontent.com/297675/29744297-4c3c357c-8a70-11e7-97e0-f67eed06ca8e.png)

CESARDELATORRE:
@infofromca Things to try:
1. Did you open the firewall ports as specified in the setup intructions?: Se section about the firewall rules. You can do it manually or running the PowerShell Script:
https://github.com/dotnet-architecture/eShopOnContainers/wiki/02.-Setting-eShopOnContainers-in-a-Visual-Studio-2017-environment 

2. Is the Identity.API project container up and running when you do a "docker ps" from command line?

CESARDELATORRE:
Also, try deleting all the old Docker images from eShopOnContainers registered in your local Docker and re-deploy.
You can do it by running this PowerShell Script:
https://github.com/dotnet-architecture/eShopOnContainers/blob/netcore2/cli-windows/delete-vs-and-eshop-images.ps1
It might be because any older Docker image was not updated.

buzzfrog:
I have the same problem. I'm on branch netcore2. I have done both things that you mentioned above. Image identity.api:dev is up and running.

CESARDELATORRE:
The DEV branch is already NET Core 2. A brand-new for clone should work for you. Make sure you delete older images, just in case.

CESARDELATORRE:
The DEV branch is already NET Core 2. A brand-new git clone should work for you. Make sure you delete older images, just in case.

huangmaoyixxx:
I have same problem and  solved it  by reading followed text in the question area

[...]

PR #1019, opened 2019-04-28
Add Powershell script to set network category to private for DockerNAT.

gababu:
Adding powershell script to set DockerNAT network category to private which solves the problem referred in https://github.com/dotnet-architecture/eShopOnContainers/issues/295 and proposal from https://github.com/dotnet/docs/issues/11528. 

This can be a workaround to run this script until docker figures out a better way to solve the issue. 
The setting is required to be done at restart, however, setting this as scheduled task to run this script might help. (I can update the documentation too once this is accepted and merged)

### DIFF
--- a/cli-windows/set-dockernat-networkategory-to-private.ps1
+++ b/cli-windows/set-dockernat-networkategory-to-private.ps1
@@ -1,0 +1,1 @@
+﻿Get-NetConnectionProfile | Where-Object { $_.InterfaceAlias -match "(DockerNAT)" } | ForEach-Object { Set-NetConnectionProfile -InterfaceIndex $_.InterfaceIndex -NetworkCategory Private }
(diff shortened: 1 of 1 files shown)


cli-windows/set-dockernat-networkategory-to-private.ps1
```diff
@@ -0,0 +1 @@
+﻿Get-NetConnectionProfile | Where-Object { $_.InterfaceAlias -match "(DockerNAT)" } | ForEach-Object { Set-NetConnectionProfile -InterfaceIndex $_.InterfaceIndex -NetworkCategory Private }
```
mvelosop: I think it'd be great to include a comment stating this needs to run in an elevated session.

mvelosop:
cc/ @CyrexCore2k

cli-windows/set-dockernat-networkategory-to-private.ps1
gababu: I have updated the script to Have #Requires -RunAsAdministrator at the start, so it provides clear message when someone executes with without elevated privilege. 

.\set-dockernat-networkategory-to-private.ps1 : The script 'set-dockernat-networkategory-to-private.ps1' cannot be run
because it contains a "#requires" statement for running as Administrator. The current Windows PowerShell session is
not running as Administrator. Start Windows PowerShell by  using the Run as Administrator option, and then try running
the script again.
At line:1 char:1
+ .\set-dockernat-networkategory-to-private.ps1
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : PermissionDenied: (set-dockernat-n...-to-private.ps1:String) [], ScriptRequiresException
    + FullyQualifiedErrorId : ScriptRequiresElevation

mvelosop:
Thanks @gababu!

Merging now 😊

PR #1036, opened 2019-05-08
UpdateAllBasketAsync should use the current basket if it exists

LeeDumond:
In the case where there is an existing basket for the current buyer, the UpdateAllBasketAsync action in the BasketController will ignore it. Instead it will always instantiate and return a new basket. This fix uses the current basket if one already exists.


LeeDumond:
Yes sure will, sorry for the delay, been so busy lately :)

mvelosop:
No worries @LeeDumond, thanks for the PR, looks good 😊

Will merge tomorrow, after the checks finish!

Cheers!

PR #1060, opened 2019-06-03
check if firewall rules exist is failing.

EdmondShtogu:
the check shall be like ~ Get-NetFirewallRule -DisplayName eShopOnContainers-* -ErrorAction Stop


sughosneo:
Hi @EdmondShtogu, please take the latest and create a new PR if the issue exists.

thank you.

PR #1092, opened 2019-07-09
Several improvement and fixes in deploy-all.sh

cmilanf:
Several fixes in this PR:

  * `--docker-password` argument was not correctly handled.
  * `--docker-username` was not correctly documented in `usage()`, `--docker-user` was incorrectly referenced instead.
  * `--dns aks` argument was not documented at usage.
  * No authentication to custom registry was performed in the script, so `--docker-username` and `--docker-password` arguments didn't have any effect. Now it is possible to push the images to custom registry.
  * `image_tag` was not aligned with `docker-compose build` output, as it will add `linux-` prefix according to the platform. A check to detect this and tag the images accordingly has been implemented.

### DIFF
--- a/k8s/helm/deploy-all.sh
+++ b/k8s/helm/deploy-all.sh
@@ -14,8 +14,9 @@ Parameters:
     The resource group for the AKS cluster. Required when the registry (using the -r parameter) is set to "aks".
   -b | --build-solution
     Force a solution build before deployment (default: false).
-  -d | --dns <dns or ip address>
-    Specifies the external DNS/ IP address of the Kubernetes cluster. 
+  -d | --dns <dns or ip address> | --dns aks
+    Specifies the external DNS/ IP address of the Kubernetes cluster.
+    If 'aks' is set as value, the DNS value is retrieved from the AKS. --aks-name and --aks-rg are needed.
     When --use-local-k8s is specified the external DNS is automatically set to localhost.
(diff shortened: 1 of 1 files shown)


eiximenis:
Thx!
@mvelosop LGTM!

cmilanf:
Added another commit with more fixes:
  * Added  `--namespace` argument, defaults to `eshop`.
  * Every Helm command (`helm install` and `helm ls` used with `helm delete`) will honor the namespace parameter. This way this script is no longer destructive with the Kubernetes cluster. Usage() has been updated accordingly.
  * Custom registry image push didn't honor `--docker-username` and `--docker-password`. A `docker login` has been added to fix the situation.
  * AKS DNS discovery was not working properly due to case sensitivity with the JMESPath to the `HTTPApplicationRouteZoneName`. A small two-entries array with iterative loop has been added ton ensure either way the AKS http routing DNS retrieval is successful. I think it would be more efficient to make a JMESPath query case insensitive, but to my knowledge it is not possible.
  * Helm charts were enabled to use Kubernetes Ingress, but the `Ingress.spec.rules.hosts[]` was not honored with the `$dns` variable. This has been fixed in the `helm install` commands and now the script deploys the Ingress as expected.

mvelosop:
Thanks @cmilanf, @eiximenis!

Merging now!

PR #1097, opened 2019-07-11
Display the type name in the log message

plneto:
Currently, the code `nameof(TH)` is printing the value TH in the message. This PR will change it to print the type name in the message e.g. `MyEventHandler`.


eiximenis:
Thanks!! :D

PR #1120, opened 2019-08-14
Update ConsentController.cs

el-titan:
Possible null ref in line 65

### DIFF
--- a/src/Services/Identity/Identity.API/Controllers/ConsentController.cs
+++ b/src/Services/Identity/Identity.API/Controllers/ConsentController.cs
@@ -62,12 +62,12 @@ namespace Microsoft.eShopOnContainers.Services.Identity.API.Controllers
             ConsentResponse response = null;
 
             // user clicked 'no' - send back the standard 'access_denied' response
-            if (model.Button == "no")
+            if (model != null && model.Button == "no")
             {
                 response = ConsentResponse.Denied;
             }
             // user clicked 'yes' - validate the data
-            else if (model.Button == "yes" && model != null)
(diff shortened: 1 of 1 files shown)


src/Services/Identity/Identity.API/Controllers/ConsentController.cs
```diff
@@ -62,12 +62,12 @@ public async Task<IActionResult> Index(ConsentInputModel model)
             ConsentResponse response = null;
 
             // user clicked 'no' - send back the standard 'access_denied' response
-            if (model.Button == "no")
+            if (model != null && model.Button == "no")
```
mvelosop: A shorter expression using language features:

```suggestion
            if (model?.Button == "no" ?? false)
```

src/Services/Identity/Identity.API/Controllers/ConsentController.cs
```diff
@@ -62,12 +62,12 @@ public async Task<IActionResult> Index(ConsentInputModel model)
...
             {
                 response = ConsentResponse.Denied;
             }
             // user clicked 'yes' - validate the data
-            else if (model.Button == "yes" && model != null)
+            else if (model != null && model.Button == "yes")
```
mvelosop: Same as before:

```suggestion
            else if (model?.Button == "yes" ?? false)
```

el-titan:
> Thanks for the PR @el-titan,
> 
> I left a couple of suggestions, however, since `model.` is used all over, it'd probably be better to do a simple :
> 
> ```cs
> if (model is null) return BadRequest();
> ```

agree with you

PR #1155, opened 2019-10-22
Fix typo: "Dockfiles" => "Dockerfiles"

jo-ninja:


mvelosop:
Thanks for the PR @jo-ninja, however all Dockerfiles were changed and moved with the latest release, so this became outdated.

PR #1170, opened 2019-11-12
Fix Image URLs on Read me file

gababu:
Fix Image URLs on Read me file. It seems to be broken


mvelosop:
Hi @gababu, thanks for the heads up and the PR.

LGTM, so merging now 😊

PR #1231, opened 2020-01-28
Use internal docker dns names

mvelosop:
Change references from `localhost` to `docker.for.win.localhost` to solve MVC login when using Docker Desktop 2.2.0.0

See the [.env file](https://github.com/dotnet-architecture/eShopOnContainers/blob/fix/docker-2.2-mvc-login/src/.env) for other platforms.

For testing!


mvelosop:
Thanks @01011011!

PR #1237, opened 2020-02-05
Overload `==` and `!=` in ValueObject

christianfosli:
I was reading [these docs](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/implement-value-objects) and found it a bit confusing that the `ValueObject` class contains the methods `EqualOperator` and `NotEqualOperator`, but it doesn't override these operators.

I suggest to override the `==` and `!=` operators instead.

If this is approved I can make a similar PR in the docs repo to update that article :smile:

### DIFF
--- a/src/Services/Ordering/Ordering.Domain/SeedWork/ValueObject.cs
+++ b/src/Services/Ordering/Ordering.Domain/SeedWork/ValueObject.cs
@@ -5,7 +5,7 @@ namespace Microsoft.eShopOnContainers.Services.Ordering.Domain.SeedWork
 {
     public abstract class ValueObject
     {
-        protected static bool EqualOperator(ValueObject left, ValueObject right)
+        public static bool operator ==(ValueObject left, ValueObject right)
         {
             if (ReferenceEquals(left, null) ^ ReferenceEquals(right, null))
             {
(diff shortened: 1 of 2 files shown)


sughosneo:
Hello @christianfosli, 

Thank you for submitting this PR and the suggestions.

If you see the [Value object implementation in C#](https://docs.microsoft.com/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/implement-value-objects#value-object-implementation-in-c) example, you would notice class `ValueObject` contains the definition of different utility methods like `EqualOperator` and `NotEqualOperator`. 

Now in `EqualOperator()` , `ReferenceEquals()` method of `System.Object` class has been used to do the comparison between two different objects and so do in the `NotEqualOperator` method, as it's internally calling the same `EqualOperator` method.

So in this specific scenario, there might be any need of doing operator overloading as the functionality is already achieved. To my opinion, it would be better if we leave these methods as is. :)

Also, recently there have been some updates on ValueObject in both `docs` and `eShopOnContainers`. You may want to check those out as well.

christianfosli:
Okay. When I think of a valueobject, I expect `==` to be true when two instances have the same values, while the current implementation requires calling those specific methods to achieve that functionality. I found that a little bit misleading. With C# 9 and records coming out soon, I think records will be a better approach to achieve this anyways.

Thanks for the review, though, I will close this 😊

PR #1260, opened 2020-03-04
Added .editorconfig

rafsanulhasan:


sughosneo:
Hello @rafsanulhasan, Thank you for submitting this PR.

The editor configurations are specific to IDE and not to the application configuration. It can be useful for individual but it's not required for an application to have as a part of it's repository. 
Particularly when an application is being actively developed by many individuals and each one of them might have their choice of editor or IDE.
So unfortunately we will not able to accept this PR. 

All the active work are being tracked under `issues`. You can pick up any of an item of your choice and send us a PR. We will be happy to review and merge in the changes.

PR #1288, opened 2020-04-04
Private readonly string changed to private const string

flarocca:


sughosneo:
Hi @flarocca, merged with local copy of the dev branch and tested it. Looks good. Changes work fine for both MVC and SPA.

Issue #1303, opened 2020-04-18
CatalogService: GetItemsByIds wrong behavior  


dmytrohridin:
Hi all,
In case when items list will be empty - Status will be set to 'NotFound', but then the value will be written to 'Ok'. Seems it's wrong behavior. If yes - PR has been created.  
![image](https://user-images.githubusercontent.com/4683157/79643371-50033680-81ab-11ea-831b-63a7395b0d02.png)

dmytrohridin:
Hello @mvelosop
Can I ask you to look at this issue? Seems it's a bug. If not - feel free to close it.
Thank you :)

sughosneo:
Hi @dmytrohridin, thank you for pointing that out. Yes, that's definitely a scope of improvement.
As the client `await GrpcCallerService.CallService(` relying on `return this.MapToResponse(` statement that's the reason it's not doing any check on `ServerCallContext` status.  The correct status is useful for the unit test and for the client as well.

Accepting the PR and closing the issue.

PR #1304, opened 2020-04-18
CatalogService: Fix issue with Status set when items list is empty

dmytrohridin:
https://github.com/dotnet-architecture/eShopOnContainers/issues/1303


amroSaad78:
I think it should returns empty list
aaa

amroSaad78:
aaa

I think now it will be ok.

PR #1316, opened 2020-05-06
Refactored ValueObject's Equals() method

andrefilipegsilva:
This PR attempts to fix this bug: https://github.com/dotnet/docs/issues/7477

Although I'm not entirely sure how can two `GetAtomicValues()` methods from the same type return a different amount of objects. The only case I am thinking on is if we for any reason yield items of a member collection, and those two collections on both instances have a different number of items.

Anyways, for any case, I fixed that problem by using LINQ's `SequenceEqual()` ext method, which guarantees both equal count and item equality, and allows us to remove a bunch of code. I tested multiple equality scenarios, still if you think I forgot an important one, please let me know.

I've also changed the method name from `GetAtomicValues` to `GetEqualityComponents`. I think the name suits better.

Up for discussion! It's my first contribution, so please be kind 😄

### DIFF
--- a/src/Services/Ordering/Ordering.Domain/AggregatesModel/OrderAggregate/Address.cs
+++ b/src/Services/Ordering/Ordering.Domain/AggregatesModel/OrderAggregate/Address.cs
@@ -23,7 +23,7 @@ namespace Microsoft.eShopOnContainers.Services.Ordering.Domain.AggregatesModel.O
             ZipCode = zipcode;
         }
 
-        protected override IEnumerable<object> GetAtomicValues()
+        protected override IEnumerable<object> GetEqualityComponents()
         {
             // Using a yield return statement to return each element one at a time
             yield return Street;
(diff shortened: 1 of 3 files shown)


rmeoc:
I like your fix. I tried to fix the same issue, see pull request #1263, but I like your idea of using SequenceEqual better because it gets rid of all the tricky code.

sughosneo:
Hi @andrefilipegsilva , thank you for submitting this PR.

Your implementation of `SequenceEqual()` extension method looks good. But in my opinion, it would be better if we keep the original name `GetAtomicValues` instead of `GetEqualityComponents`. Because if you look into the `Address` class example, you would notice method `GetAtomicValues()` yields each element one at a time. Though there is nothing wrong with the name
`GetEqualityComponents` but `GetAtomicValues` this seems to be more aligned in this context. 

So if you could make the necessary changes in both `ValueObject.cs` and `ValueObjectTests.cs`. Will test and merge the changes in `dev` branch. 

Please feel free to share your thoughts.

andrefilipegsilva:
Hey @sughosneo

Thanks for your input and for updating the value object doc with my implementation!

However, I would still change the name to `GetEqualityComponents() `. In my opinion, `GetAtomicValues()` indicates that I will be returning a collection of values that compose a specific Value Object. I think we should not default this behavior for equality purposes, because not all atomic members may actually be equality components.

sughosneo:
Hi @andrefilipegsilva, We reviewed it again and we think its good to go with the current changes. Merging this PR.
Also, please note we updated our docs with this [PR-20235](https://github.com/dotnet/docs/pull/20235)

Thank you for your feedback !

PR #1344, opened 2020-06-17
Obsolete

william-keller:
Obsolete SwaggerGenOptions.DescribeAllEnumsAsStrings()


mvelosop:
Thanks @william-keller,

However, I think that the [Obsolete] attribute is mostly for library use, so that the author can inform developers that some API is going to be deprecated and give some related information like the replacing API.

Since this is not a library, but just app code, the proper way to handle this would be to replace the obsolete function usage.

PR #1349, opened 2020-06-23
Fix RemoveBackStackAsync in NavigationService (Xamarin)

### LINKED ISSUE #1210: Unexpected behavior in RemoveBackStackAsync method (NavigationService.cs)
In the RemoveBackStackAsync method the following expression is evaluated before every loop iteration `i < mainPage.Navigation.NavigationStack.Count - 1`, and the `NavigationStack.Count` will change, right?

This means that not all the previous items in the stack will be removed.
For example in a stack of fourth items only two items will be removed

```
        public Task RemoveBackStackAsync()
        {
            var mainPage = Application.Current.MainPage as CustomNavigationView;

            if (mainPage != null)
            {
                for (int i = 0; i < mainPage.Navigation.NavigationStack.Count - 1; i++)
                {
                    var page = mainPage.Navigation.NavigationStack[i];
                    mainPage.Navigation.RemovePage(page);
                }
            }

            return Task.FromResult(true);
        }
```

I changed the previous method to the following

```
        public Task RemoveBackStackAsync()
        {
            var mainPage = Application.Current.MainPage as CustomNavigationView;

            if (mainPage != null)
            {
		var navigationCount = mainPage.Navigation.NavigationStack.Count;
                for (int i = 0; i < navigationCount - 1; i++)
                {
                    var page = mainPage.Navigation.NavigationStack[0];
                    mainPage.Navigation.RemovePage(page);
                }
            }

            return Task.FromResult(true);
        }
```
Maybe I'm missing something but I had to change the implementation for the second approach.

Miguel1422:
This fixes #1210

### DIFF
--- a/src/Mobile/eShopOnContainers/eShopOnContainers.Core/Services/Navigation/NavigationService.cs
+++ b/src/Mobile/eShopOnContainers/eShopOnContainers.Core/Services/Navigation/NavigationService.cs
@@ -66,7 +66,7 @@ namespace eShopOnContainers.Services
 
             if (mainPage != null)
             {
-                for (int i = 0; i < mainPage.Navigation.NavigationStack.Count - 1; i++)
+                for (int i = mainPage.Navigation.NavigationStack.Count - 2; i >= 0; i--)
                 {
                     var page = mainPage.Navigation.NavigationStack[i];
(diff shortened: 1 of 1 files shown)


jamesmontemagno:
Removing this as we moved the repo and is now .NET MAUI

PR #1355, opened 2020-07-02
Swamy/remove unused using and remove unused httpclient

vishipayyallore:
1. Removed Unused using
2. The single file has two classes, I have moved into a separate file.
3. Added line space between the Properties of the class for readability. 
4. Removed the HttpClient from 3 files.


vishipayyallore:
Duplicate item

PR #1357, opened 2020-07-03
Swamy/remove unused using and remove unused httpclient

vishipayyallore:
1. Removed unused using statements. Also sorted using statements.
2. Added line spaces between the properties. Also moved classes into separate files. Some files have two classes inside.
3. Removed unused HTTP Client from 3 files. Those 3 files were using gRPC communication.

### DIFF
--- a/src/ApiGateways/Web.Bff.Shopping/aggregator/Config/UrlsConfig.cs
+++ b/src/ApiGateways/Web.Bff.Shopping/aggregator/Config/UrlsConfig.cs
@@ -1,7 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
+﻿using System.Collections.Generic;
 
 namespace Microsoft.eShopOnContainers.Web.Shopping.HttpAggregator.Config
 {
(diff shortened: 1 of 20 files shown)


sughosneo:
Thanks @vishipayyallore for submitting this PR. 

HttpClient gets injected within the public constructor of these services CatalogService, BasketService and OrderApiClient and needs to be there because those services have been registered in *Startup.cs* class as AddHttpClient(). For e.g

```C#
services.AddHttpClient<IBasketService, BasketService>()
                .AddHttpMessageHandler<HttpClientAuthorizationDelegatingHandler>()
                .AddDevspacesSupport();
```
If we remove _httpClient from the Basket service it may throw error while user would try add an item into a cart. So we wouldn't be able to merge these changes.

However this PR also contains other useful changes, which we would like to incorporate. For e.g :

	a) Unused using statement in different classes.
	b) Refactoring classes in a separate PR

Could you please open up separate PRs on those items ? It would be easy for us to merge those items separately.

Also, we see there is an another open PR #1355 on the same items with a different branch name. You may want to close that too while creating new ones. 

Thanks again.

vishipayyallore:
@sughosneo 
Sure I will close this and open another PR with below mentioned two items.

a) Unused using statement in different classes.
b) Refactoring classes in a separate PR

PR #1416, opened 2020-08-30
Remove AddOrderItem method from Order class

AndriiTokarskyi:
Remove AddOrderItem from Order's public interface since it is not its behaviour.

### DIFF
--- a/src/Services/Ordering/Ordering.API/Application/Commands/CreateOrderCommandHandler.cs
+++ b/src/Services/Ordering/Ordering.API/Application/Commands/CreateOrderCommandHandler.cs
@@ -1,4 +1,6 @@
-﻿namespace Microsoft.eShopOnContainers.Services.Ordering.API.Application.Commands
+﻿using System.Linq;
+
+namespace Microsoft.eShopOnContainers.Services.Ordering.API.Application.Commands
 {
     using Domain.AggregatesModel.OrderAggregate;
     using global::Ordering.API.Application.IntegrationEvents;
@@ -46,12 +48,19 @@
             // methods and constructor so validations, invariants and business logic 
             // make sure that consistency is preserved across the whole aggregate
             var address = new Address(message.Street, message.City, message.State, message.Country, message.ZipCode);
-            var order = new Order(message.UserId, message.UserName, address, message.CardTypeId, message.CardNumber, message.CardSecurityNumber, message.CardHolderName, message.CardExpiration);
+            var orderItems = message.OrderItems.Select(item => new OrderItem(item.ProductId, item.ProductName,
+                item.UnitPrice, item.Discount, item.PictureUrl, item.Units));
 
-            foreach (var item in message.OrderItems)
-            {
-                order.AddOrderItem(item.ProductId, item.ProductName, item.UnitPrice, item.Discount, item.PictureUrl, item.Units);
-            }
+            var order = new Order(
+                message.UserId,
+                message.UserName,
+                address,
+                message.CardTypeId,
+                message.CardNumber,
+                message.CardSecurityNumber,
+                message.CardHolderName,
+                message.CardExpiration,
+                orderItems.ToList());
 
             _logger.LogInformation("----- Creating Order - Order: {@Order}", order);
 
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet-architecture/eShopOnContainers?pullRequest=1416) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet-architecture/eShopOnContainers?pullRequest=1416) before we can accept your contribution.<br/><br/>:x: AndriiTokarskyi [sign now](https://cla.dotnetfoundation.org/dotnet-architecture/eShopOnContainers?pullRequest=1416)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet-architecture/eShopOnContainers?pullRequest=1416) it.</sub>

pavanarya:
@AndriiTokarskyi AndriiTokarskyi I have a quick question. I think we should interact with entities in a aggregate via Aggregateroot. In this case since Order is Aggregate, So I think it makes sense to have AddOrderItem in Order class. What are your thoughts?

maranmaran:
> @AndriiTokarskyi AndriiTokarskyi I have a quick question. I think we should interact with entities in a aggregate via Aggregateroot. In this case since Order is Aggregate, So I think it makes sense to have AddOrderItem in Order class. What are your thoughts?

Yes, I believe that's the point of DDD pattern showcase within Ordering service with complex domain entities encapsulating business knowledge, validation, and other domain-related responsibilities within its aggregate root. 

https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/net-core-microservice-domain-model

`AddOrderItem` encapsulated domain rules/logic for adding the OrderItem to the order and we would expose the items as read-only collections making them immutable and enforcing rules on each item added.

Within this PR as I see, the primary driver is the ability to propagate orders to ctor and allow it to add items on its own.
This feature can be done without removing the `AddOrderItem` from the public API. 

However, I would still argue it's a poorer practice as we might want to return some validation results, etc success, errors whatever objects for each individual order item to asses it within our orchestration logic. 

For resolving bulk orders I would much rather opt into implementing AddOrderItems instead of cramming these 'possibly expanded in future' essentials into ctor. 

One more thing is that some items within this PR go out of the scope of the proposed change. (Some method renames etc)

src/Services/Ordering/Ordering.API/Application/Commands/CreateOrderCommandHandler.cs
```diff
@@ -46,12 +48,19 @@ public async Task<bool> Handle(CreateOrderCommand message, CancellationToken can
...
             var address = new Address(message.Street, message.City, message.State, message.Country, message.ZipCode);
-            var order = new Order(message.UserId, message.UserName, address, message.CardTypeId, message.CardNumber, message.CardSecurityNumber, message.CardHolderName, message.CardExpiration);
+            var orderItems = message.OrderItems.Select(item => new OrderItem(item.ProductId, item.ProductName,
+                item.UnitPrice, item.Discount, item.PictureUrl, item.Units));
 
-            foreach (var item in message.OrderItems)
-            {
-                order.AddOrderItem(item.ProductId, item.ProductName, item.UnitPrice, item.Discount, item.PictureUrl, item.Units);
-            }
+            var order = new Order(
```
maranmaran, line 54: If you wish to handle bulk complexity I would rather opt into implementing `AddOrderItems` instead of CTOR where you lose some flexibility like returning results, async, etc..

Reasons being AddOrderItem can/could change from `void` to some functional object, validation, error result with some success flag for instance since it should encapsulate some logic which currently is not there but this is still a showcase of the pattern.

src/Services/Ordering/Ordering.Domain/AggregatesModel/OrderAggregate/OrderItem.cs
```diff
@@ -43,7 +43,7 @@ public OrderItem(int productId, string productName, decimal unitPrice, decimal d
 
         public string GetPictureUri() => _pictureUrl;
 
-        public decimal GetCurrentDiscount()
+        public decimal GetDiscount()
```
maranmaran, line 46: Rename is out of proposed scope of this PR

maranmaran (changes requested):

PR #1479, opened 2020-09-28
Enable TLS for local development with docker compose

mvelosop:
Hi @nishanil, @sughosneo 

Long time no see 😉

Could take a look at this PR?

- Enables https all over for docker-compose
- Should solve most login issues originating from using http

Cheers 😊


sughosneo:
> Hi @nishanil, @sughosneo
> 
> Long time no see 😉
> 
> Could take a look at this PR?
> 
> * Enables https all over for docker-compose
> * Should solve most login issues originating from using http
> 
> Cheers 😊

Hey @mvelosop , yes long time 😊. 

Thank you for submitting the PR. Will test and update.

mvelosop:
BTW, I forgot to mention that the start procedure changes a bit, because there's the `src/docker-compose.certificates.yml` file with the certificate password, that I opted to keep apart and .gitignore'd to protect the "secrets" from going accidentally into the repo.

I added the `src/start.ps1` file to simplify that and a couple of features that I've come to use quite often when starting up eShop.

Cheers 😊

sughosneo:
> BTW, I forgot to mention that the start procedure changes a bit, because there's the `src/docker-compose.certificates.yml` file with the certificate password, that I opted to keep apart and .gitignore'd to protect the "secrets" from going accidentally into the repo.
> 
> I added the `src/start.ps1` file to simplify that and add a couple of features that I've come to use quite often when starting up eShop:
> 
> * Starting the whole app (`.\start.ps1`)
> * Starting infrastructure containers (`.\start.ps1 infra`)
> * Starting a specific list of services (`.\start.ps1 ordering-api identity-api ...`)
> 
> You also need to create the self-signed certificates, as per the `deploy\certificates\README.md`, before building the images with docker-compose.
> 
> Cheers 😊

Thank you for the update @mvelosop and sharing further details 😊
We are evaluating a few aspects at our end to see if we can keep the SSL enablement process seamless for the users in both dev/local (docker-compose & K8s) and production scenarios. That's the reason it may take a little longer to review/test than expected. Will keep you posted with an update.

mvelosop:
Thanks for the info @sughosneo,

It should work as long as you:

1. Assign a DNS name to the AKS load balancer IP
2. Add that DNS name to the self-signed certificate and
3. Import the self-signed certificate into each user's root CA store.

Cheers 😊

sughosneo:
> Thanks for the info @sughosneo,
> 
> It should work as long as you:
> 
> 1. Assign a DNS name to the AKS load balancer IP
> 2. Add that DNS name to the self-signed certificate and
> 3. Import the self-signed certificate into each user's root CA store.
> 
> Cheers 😊

Hi @mvelosop, 

I have performed a quick test of this PR. Couple of things I have noticed. 

1) The docker-compose certificate file has different extensions that what is expected. 
    For e.g : in *Start.ps1* has `docker-compose.certificates.yml`, where in the provided template has `src\docker-compose.certificates.yaml` because of that user may get the following error :

image

2) Once I change the extensions of that file (docker-compose.certificates) from `yaml` to `yml` and rerun the step all the services come up fine. 

But after that when I cleaned up all the images and try to rebuild the images again. It throws below error :

image

Am I missing anything ? Thoughts ?

Please note, I am yet to test out the entire functionality in details. I will perform further testing and try to capture more information.

Thank you.

Cheers 😊

mvelosop:
Well, there was another bug, the last line of `install-docker-certificate.ps1` should've change the certificate extension from .pem to .crt.

Pushing another commit with both fixes now.

MylesRip:
Hi @mvelosop. First, thank you for taking the initiative to add this feature!  

I tried it out and got an error when running the install-docker-certificate.ps1 file. The error is on line 18 and states, "Copy-Item : Could not find a part of the path". This happens because the repo, in its pristine state, has no "certificates" folder under "src". The fix is to add the following line just before the Copy-Item line: 

New-Item -ItemType File -Path ..\..\src\certificates\docker-self-signed.crt  -Force

This creates the directory along with an empty file which gets overwritten by the Copy-Item command that follows it.

MylesRip:
Hi again @mvelosop. 

Just a minor typo in the Readme.md file. On line 38, ".yaml" should be ".yml".  

Thank you!

PR #1494, opened 2020-10-19
Fixed parameter bug create resource script

### LINKED ISSUE #1341: bug in createresources.sh script
In Line#54 & 55 of the file createresources.sh, it should be $1, instead of $2

sughosneo:
Fixes  #1341


sughosneo:
Tested this script from `WSL` with sql and redis resource creation. For e.g :

`bash ../createresources.sh redisdeploy eShop -c westus`

PR #1520, opened 2020-11-16
Fixed #1518 bug

### LINKED ISSUE #1518: Migration failed to add a field to the CatalogItem entity in CatalogContext of Catalog.API
**1.First, I added a String field: Tests**

`public string Tests { get; set; }`

**2.Then, to perform the migration:**

```
 add - migration AddTests - c CatalogContext - o Infrastructure/CatalogMigrations
```

**3.Finally, execute the project(or update-database) and report an error**

image

------------

**After Debugging the source code, I found that the last migration record was not synchronized into the model snapshot.**

This submission record has been modified.
(https://github.com/dotnet-architecture/eShopOnContainers/commit/45ed8a60ad87c0da8f84f93aae143dce7e9c8bcc)
But the snapshot file has not changed.
（https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/src/Services/Catalog/Catalog.API/Infrastructure/CatalogMigrations/CatalogContextModelSnapshot.cs）

Causes new fields to be added to conflict.

anjoy8:
Update CatalogContextModelSnapshot.cs

### DIFF
--- a/src/Services/Catalog/Catalog.API/Infrastructure/CatalogContextSeed.cs
+++ b/src/Services/Catalog/Catalog.API/Infrastructure/CatalogContextSeed.cs
@@ -234,7 +234,7 @@
                 Description = column[Array.IndexOf(headers, "description")].Trim('"').Trim(),
                 Name = column[Array.IndexOf(headers, "name")].Trim('"').Trim(),
                 Price = price,
-                PictureUri = column[Array.IndexOf(headers, "pictureuri")].Trim('"').Trim(),
+                PictureUri = column[Array.IndexOf(headers, "picturefilename")].Trim('"').Trim(),
             };
 
             int availableStockIndex = Array.IndexOf(headers, "availablestock");
--- a/src/Services/Catalog/Catalog.API/Infrastructure/CatalogContextSeed.cs
+++ b/src/Services/Catalog/Catalog.API/Infrastructure/CatalogContextSeed.cs
@@ -181,7 +181,7 @@
             string[] csvheaders;
             try
             {
-                string[] requiredHeaders = { "catalogtypename", "catalogbrandname", "description", "name", "price", "pictureFileName" };
+                string[] requiredHeaders = { "catalogtypename", "catalogbrandname", "description", "name", "price", "picturefilename" };
                 string[] optionalheaders = { "availablestock", "restockthreshold", "maxstockthreshold", "onreorder" };
                 csvheaders = GetHeaders(csvFileCatalogItems, requiredHeaders, optionalheaders );
             }
--- a/src/Services/Catalog/Catalog.API/Infrastructure/CatalogContextSeed.cs
+++ b/src/Services/Catalog/Catalog.API/Infrastructure/CatalogContextSeed.cs
@@ -234,7 +234,7 @@
                 Description = column[Array.IndexOf(headers, "description")].Trim('"').Trim(),
                 Name = column[Array.IndexOf(headers, "name")].Trim('"').Trim(),
                 Price = price,
-                PictureUri = column[Array.IndexOf(headers, "picturefilename")].Trim('"').Trim(),
+                PictureFileName = column[Array.IndexOf(headers, "picturefilename")].Trim('"').Trim(),
             };
 
             int availableStockIndex = Array.IndexOf(headers, "availablestock");
--- a/src/Services/Catalog/Catalog.API/Infrastructure/CatalogMigrations/CatalogContextModelSnapshot.cs
+++ b/src/Services/Catalog/Catalog.API/Infrastructure/CatalogMigrations/CatalogContextModelSnapshot.cs
@@ -14,9 +14,10 @@ namespace Catalog.API.Infrastructure.Migrations
         {
             modelBuilder
                 .HasAnnotation("ProductVersion", "1.1.1")
-                .HasAnnotation("SqlServer:Sequence:.catalog_brand_hilo", "'catalog_brand_hilo', '', '1', '10', '', '', 'Int64', 'False'")
-                .HasAnnotation("SqlServer:Sequence:.catalog_hilo", "'catalog_hilo', '', '1', '10', '', '', 'Int64', 'False'")
-                .HasAnnotation("SqlServer:Sequence:.catalog_type_hilo", "'catalog_type_hilo', '', '1', '10', '', '', 'Int64', 'False'")
+                .HasAnnotation("Relational:MaxIdentifierLength", 128)
+                .HasAnnotation("Relational:Sequence:.catalog_brand_hilo", "'catalog_brand_hilo', '', '1', '10', '', '', 'Int64', 'False'")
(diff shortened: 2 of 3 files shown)


sughosneo:
Hi @anjoy8 , thank you for submitting this PR. Will test and update.

sughosneo:
> Hi @anjoy8 , thank you for submitting this PR. Will test and update.

Hi @anjoy8, I did a quick test and I could see that it didn't update the Catalog table. Could you please test and debug it at your end ? You can follow below steps to reproduce the behavior :

To clean up :

```powershell
docker stop $(docker ps -aq)
docker rm $(docker ps -aq)
docker volume rm $(docker volume ls -q)
```
>!Note : Above commands will clean up all your docker setup. It's expected to run it only on `dev` environment. 

And to run :

```powershell
docker-compose up sqldata
docker-compose up
```

I will review the code in details.

Thank you.

anjoy8:
**Sorry, I'm fine here, but I'm not using Docker-compose, 
normal debugging and running.**

image

image


image

sughosneo:
Hi @anjoy8, thanks for the update. 

You should be able to reproduce the issue by the following steps :

- Perform clean up (mentioned above)
- Set the `docker-compose` project as a startup project in your Visual Studio environment.
- Then press `F5` to run the solution.
- You can place a break point `CatalogItem CreateCatalogItem` method to debug.
- Or once all the containers are up and running you would notice that though `[CatalogBrand]` and `[CatalogType]` tables are populated but `[dbo].[Catalog]` table is still empty. 

The reason it's not populating `CatalogItem` because it's throwing exceptions for line 203 in class `CatalogContextSeed` which actually gets originated from line `PictureUri = column[Array.IndexOf(headers, "pictureuri")].Trim('"').Trim()` within the `CreateCatalogItem()` method. That needs to be updated to incorporate `PictureFileName`. For e.g :`PictureFileName = column[Array.IndexOf(headers, "picturefilename")].Trim('"').Trim()`

>!Note if you are connecting to a different SQL data source than `docker` you may have to drop the database `Microsoft.eShopOnContainers.Services.CatalogDb` to let it get populated again. 

Now when you select `Catalog.API` as a startup project and run that (*let's say after brining up only the sql container by using `docker-compose up sqldata`*), then you would notice the control doesn't go to the `GetCatalogItemsFromFile()` method. Actually the data gets populated from the `GetPreconfiguredItems()` method. That's the reason in your local you see everything is populated.

You may want to fix this entire flow. And once you make the necessary changes you can test out the following scenarios just to double confirm everything is working fine. 

- Visual Studio F5
- docker-compose up
- local k8s

Hope this helps !

anjoy8:
Hi @sughosneo , thanks for your suggestion and reply!
After my debug, I found that there was something wrong with the CSV data, 
and I submitted a PR(https://github.com/dotnet-architecture/eShopOnContainers/pull/1520/commits/ad8ec3f5225a84d94688a9710a975e817827af6a). You can test it again. 
If there is a problem, Please let me know the error message.
I will change it.

src/Services/Catalog/Catalog.API/Infrastructure/CatalogContextSeed.cs
```diff
@@ -234,7 +234,7 @@ private CatalogItem CreateCatalogItem(string[] column, string[] headers, Diction
                 Description = column[Array.IndexOf(headers, "description")].Trim('"').Trim(),
                 Name = column[Array.IndexOf(headers, "name")].Trim('"').Trim(),
                 Price = price,
-                PictureUri = column[Array.IndexOf(headers, "pictureuri")].Trim('"').Trim(),
+                PictureUri = column[Array.IndexOf(headers, "picturefilename")].Trim('"').Trim(),
```
sughosneo, line 237: Wouldn't this be better to rename this variable as well ? For e.g 
`PictureFileName = column[Array.IndexOf(headers, "picturefilename")].Trim('"').Trim()`

sughosneo:
> Hi @sughosneo , thanks for your suggestion and reply!
> After my debug, I found that there was something wrong with the CSV data,
> and I submitted a PR([ad8ec3f](https://github.com/dotnet-architecture/eShopOnContainers/commit/ad8ec3f5225a84d94688a9710a975e817827af6a)). You can test it again.
> If there is a problem, Please let me know the error message.
> I will change it.

Thanks for the update @anjoy8 . I have left a quick review comment. 
Also, wouldn't this be just be considered as name `Roslyn Red T-Shirt"` with `"` character ? 

Does it work in `docker-compose` scenario as well ?

src/Services/Catalog/Catalog.API/Infrastructure/CatalogContextSeed.cs
anjoy8: yes,you are right，I rename this variable

anjoy8:
> 
> 
> > Hi @sughosneo , thanks for your suggestion and reply!
> > After my debug, I found that there was something wrong with the CSV data,
> > and I submitted a PR([ad8ec3f](https://github.com/dotnet-architecture/eShopOnContainers/commit/ad8ec3f5225a84d94688a9710a975e817827af6a)). You can test it again.
> > If there is a problem, Please let me know the error message.
> > I will change it.
> 
> Thanks for the update @anjoy8 . I have left a quick review comment.
> Also, wouldn't this be just be considered as name `Roslyn Red T-Shirt"` with `"` character ?
> 
> Does it work with `docker-compose` scenario as well ?

**But here Split is " character  , so it will truncate the data,**
图片


**An exception occurs:**
图片

**Although this does not affect the operation of the project.**

*******

```
Also, sorry, I didn't use the Docker-compose test, if I still report an error.
You can modify it on your machine according to my modification ideas. I can turn off this PR.
```

sughosneo:
> > > Hi @sughosneo , thanks for your suggestion and reply!
> > > After my debug, I found that there was something wrong with the CSV data,
> > > and I submitted a PR([ad8ec3f](https://github.com/dotnet-architecture/eShopOnContainers/commit/ad8ec3f5225a84d94688a9710a975e817827af6a)). You can test it again.
> > > If there is a problem, Please let me know the error message.
> > > I will change it.
> > 
> > 
> > Thanks for the update @anjoy8 . I have left a quick review comment.
> > Also, wouldn't this be just be considered as name `Roslyn Red T-Shirt"` with `"` character ?
> > Does it work with `docker-compose` scenario as well ?
> 
> **But here Split is " character , so it will truncate the data,**
> 
> `.Select(row => Regex.Split(row, ",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)") )`
> 
> 图片
> 
> **An exception occurs:**
> 图片
> 
> **Although this does not affect the operation of the project.**
> 
> ```
> Also, sorry, I didn't use the Docker-compose test, if It still report an error.
> You can modify it on your machine according to my modification ideas. I can turn off this PR.
> ```

Thanks @anjoy8 . Will test and update.

sughosneo:
I have tested this PR with `docker-compose` and `local-k8s` scenario. It works fine. So merging the changes.

Thanks @anjoy8

PR #1525, opened 2020-11-23
Fix PurchaseUrl port in WebSPA appsettings.json

n-stefan:
Hi,
in the WebSPA project's appsettings.json the port of PurchaseUrl seems to be wrong but it doesn't cause errors because the matching environment variable in docker-compose.override.yml overrides it.
Commenting out the environment variable results in the catalog not being shown and console errors like:

Access to XMLHttpRequest at 'http://localhost:5200/c/api/v1/catalog/items?pageIndex=0&pageSize=10' from origin 'http://host.docker.internal:5104' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

The change in this PR fixes it.

### DIFF
--- a/src/Web/WebSPA/appsettings.json
+++ b/src/Web/WebSPA/appsettings.json
@@ -2,7 +2,7 @@
   "IdentityUrl": "http://localhost:5105",
   "MarketingUrl": "http://localhost:5110",
   "CallBackUrl": "http://localhost:5104/",
-  "PurchaseUrl": "http://localhost:5200",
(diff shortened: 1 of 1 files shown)


sughosneo:
Hi @n-stefan, thank you for submitting this PR. Good catch 👍 

Yes, the value of `PurchaseUrl` is not correct in `appsettings.json` file.  
No breaking changes. Looks good. Accepting this PR.

PR #1565, opened 2020-12-28
replace azure pipelines with github actions

borjasanes:
- Github actions
- Remove azure pipelines


mvelosop (changes requested):
Thanks @borjasanes 😊

Everything looks fine, the only comment is that we can eliminate completely the Windows-related jobs.

mvelosop:
Thanks @borjasanes.

LGTM so merging now!

mvelosop (approved):

PR #1566, opened 2020-12-28
Updated to aspnet 5 dockerfiles.development for azds

borjasanes:


src/Services/Ordering/Ordering.API/Dockerfile.develop
```diff
@@ -13,10 +13,11 @@ COPY ["BuildingBlocks/WebHostCustomization/WebHost.Customization/WebHost.Customi
 COPY ["Services/Ordering/Ordering.Domain/Ordering.Domain.csproj", "Services/Ordering/Ordering.Domain/"]
 COPY ["Services/Ordering/Ordering.Infrastructure/Ordering.Infrastructure.csproj", "Services/Ordering/Ordering.Infrastructure/"]
 COPY ["Services/Ordering/Ordering.API/Ordering.API.csproj", "Services/Ordering/Ordering.API/"]
+COPY ["NuGet.config", "NuGet.config"]
 
 RUN dotnet restore Services/Ordering/Ordering.API/Ordering.API.csproj
 COPY . .
 WORKDIR /src/Services/Ordering/Ordering.API
-RUN dotnet build --no-restore -c $BUILD_CONFIGURATION
+RUN dotnet build -c $BUILD_CONFIGURATION
```
mvelosop, line 21: Was it necessary to remove the `--no-restore`? 

It would be (or at least was) redundant, as there's a `dotnet restore` right before the build

mvelosop (changes requested):
Thanks @borjasanes 😊

There's just a minor comment that could have to be considered for all files in case it has to be changed.

src/Services/Ordering/Ordering.API/Dockerfile.develop
borjasanes: After updating to .NET 5 I had to remove it otherwise I was having problems to restore all packages.

mvelosop: Great, thanks!

mvelosop (approved):

PR #1584, opened 2021-01-20
signalrhub k8s setting identity

borjasanes:
ordering signalr api connect to identity via k8s network and not localhost


mvelosop:
Thanks @borjasanes, merging now!

PR #1591, opened 2021-01-24
fix envoy config in helm

borjasanes:
image


mvelosop:
Thanks @borjasanes, merging now

PR #1595, opened 2021-01-26
ordering signalr to top level statement

borjasanes:


mvelosop:
Thanks @borjasanes, merging now 😊

PR #1596, opened 2021-01-26
workflow path filter including subfolders

borjasanes:


mvelosop:
Thanks @borjasanes, merging now 😊

PR #1637, opened 2021-03-23
Fix WebSPA catalog filter

n-stefan:
Hi,

if in the catalog you
1. click "Next"
2. select ".NET" from the "Brand" drop-down
3. click ">" to filter the products

WebMVC and WebSPA show different results:

WebMVC: "Showing 6 of 6 products - Page 1 - 1" (goes back to first page)
WebSPA: "THERE ARE NO RESULTS THAT MATCH YOUR SEARCH" (stays on second page)

Assuming WebMVC's behavior is the desired one, this PR changes WebSPA to do the same.


sughosneo (approved):
Hi @n-stefan , thank you for submitting this PR. Looks good.

PR #1663, opened 2021-05-05
Enhance Open Telemetry specific features

sughosneo:
#### This PR includes following changes  : 

- Refactored OpenTelemetry related configuration into a different project `OpenTelemetry.Customization.csproj`
- Included `Zipkin` Helm chart
- Enabled default trace with `Zipkin` in K8s deployment. 
- Enabled `Zipkin` related configuration for all the other services. 

##### docker-compose : 

```powershell
docker-compose -f docker-compose.yml -f docker-compose.override.yml -f docker-compose.opentelemetry.zipkin.yml up
```

##### local kubernetes : 

```powershell
.\deploy-all.ps1 -imageTag linux-latest -useLocalk8s $true -imagePullPolicy Never -enableTrace $true
```


src/OpenTelemetry/Extensions/OpenTelemetryExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+                    {
+                        options.Endpoint = new Uri(openTelemetryConfig.ExportToolEndpoint);
+                    });
+                    break;
+                default:
+                    tracerProviderBuilder.AddConsoleExporter();
+                    break;
+            }
+
+            tracerProviderBuilder.Build();
```
cijothomas: If the provider being built here is not stored, then it cannot be disposed.

src/OpenTelemetry/OpenTelemetry.Customization.csproj
```diff
@@ -0,0 +1,22 @@
...
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="StackExchange.Redis" Version="2.2.4" />
+		<PackageReference Include="OpenTelemetry" Version="1.0.1" />
+		<PackageReference Include="OpenTelemetry.Exporter.Console" Version="1.0.1" />
+		<PackageReference Include="OpenTelemetry.Exporter.Jaeger" Version="1.0.1" />
+		<PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.0.1" />
+		<PackageReference Include="OpenTelemetry.Exporter.Zipkin" Version="1.0.1" />
+		<PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.0.0-rc2" />
```
cijothomas, line 14: It'd be best to not use Extensions.hosting package today, and stick with manual config. This package is expected to undergo more changes.

src/OpenTelemetry/Extensions/OpenTelemetryExtensions.cs
sughosneo: Thank you @cijothomas, for the review and sorry for the delay in updating the changes. 
Yes, updated the file `src/OpenTelemetry/Extensions/OpenTelemetryExtensions.cs` and used the `tracerProviderBuilder` as local parameter.

src/OpenTelemetry/OpenTelemetry.Customization.csproj
sughosneo: ok, sure. 
It's being used for the `AddOpenTelemetryTracing()` method.  Are we supposed to use any other package? Because I see the same package is being used in this [example](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/AspNetCore/Examples.AspNetCore.csproj) as well.

velosojose:
hi

PR #1672, opened 2021-05-13
Updates the postcss package version

sughosneo:
This PR updates the `postcss` package version to `8.2.10`


sughosneo:
Tested the changes both in Windows OS and MacOS. It's working fine.

PR #1708, opened 2021-06-23
Local K8s change DNS to host.docker.internal

maucaro:
Changing the dns entry in the script for local deployments from 'localhost' to 'host.docker.internal' allows for a clean local install without the need for the section "Known Behaviors" at [https://github.com/dotnet-architecture/eShopOnContainers/wiki/Deploy-to-Local-Kubernetes#Known-Behaviours](https://github.com/dotnet-architecture/eShopOnContainers/wiki/Deploy-to-Local-Kubernetes#Known-Behaviours).

Not only is it cleaner and faster but also less error prone. The instructions in said section did not work for me as the SQL Server container presented errors after restarting it (failure to login as 'sa').

Note that 'host.docker.internal' is specific to Docker Desktop but this should be ok as the document is specific to that environment.

### DIFF
--- a/deploy/k8s/helm/deploy-all.sh
+++ b/deploy/k8s/helm/deploy-all.sh
@@ -160,7 +160,7 @@ ingress_values_file="ingress_values.yaml"
 
 if [[ $use_local_k8s ]]; then
   ingress_values_file="ingress_values_dockerk8s.yaml"
-  dns="localhost"
+  dns="host.minikube.internal"
 fi
 
 if [[ $dns == "aks" ]]; then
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet-architecture/eShopOnContainers?pullRequest=1708) <br/>All CLA requirements met.

sughosneo (commented):
Hi @maucaro, thank you for submitting this PR.
Though as of now, only Docker Desktop is only the supported environment, it might not be a good idea to hard code the DNS name to any specific tool.

Rest of the changes, we can test and merge.

deploy/k8s/helm/deploy-all.sh
```diff
@@ -160,7 +160,7 @@ ingress_values_file="ingress_values.yaml"
 
 if [[ $use_local_k8s ]]; then
   ingress_values_file="ingress_values_dockerk8s.yaml"
-  dns="localhost"
+  dns="host.minikube.internal"
```
sughosneo: Though as of now, only Docker Desktop is only the supported environment, it might not be a good idea to hard code the DNS name to any specific tool.

maucaro:
If you are only supporting Docker Desktop for local development, I would much rather have a clean experience with it than having to patch as per "known behaviors" as already cited. If you add support for others (minikube for example), you could have special cases for those as well. You already have special-case treatment for AKS, so this would not be much different. I can tell you that the current documented patching in "known behaviors" does NOT consistently work; this does. Anyway, up to you....

maucaro:
closing this pull request.

PR #1725, opened 2021-07-20
Updates ingress object version to use v1 api

sughosneo:
This PR updates the following component : 

- Removes the archived k8s directory
- `Ingress` object to API version to `networking.k8s.io/v1`


nishanil (approved):

sughosneo:
Tested the changes in the following environment. 

- Windows OS
- MacOS
- AKS

PR #1732, opened 2021-07-30
Serialize properties in derived classes 

michaelgregson:
System.Text.Json serializes based on static type of value parameter. Polymorphic serialization requires type information


sughosneo (approved):
Hi @michaelgregson, thank you for submitting this PR.
Tested locally. LGTM!

PR #1756, opened 2021-09-11
Veysel MUTLU changes

vmutlu:
## :pushpin: Changes
Some abbreviations and class additions were made to the project.

### DIFF
--- a/src/Services/Catalog/Catalog.API/Controllers/CatalogController.cs
+++ b/src/Services/Catalog/Catalog.API/Controllers/CatalogController.cs
@@ -304,10 +304,7 @@ namespace Microsoft.eShopOnContainers.Services.Catalog.API.Controllers
             var baseUri = _settings.PicBaseUrl;
             var azureStorageEnabled = _settings.AzureStorageEnabled;
 
-            foreach (var item in items)
-            {
-                item.FillProductUrl(baseUri, azureStorageEnabled: azureStorageEnabled);
-            }
+            items.ForEach(item => { item.FillProductUrl(baseUri, azureStorageEnabled: azureStorageEnabled); });
 
             return items;
         }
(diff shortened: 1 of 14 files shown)


src/Services/Catalog/Catalog.API/Controllers/CatalogController.cs
```diff
@@ -304,10 +304,7 @@ private List<CatalogItem> ChangeUriPlaceholder(List<CatalogItem> items)
             var baseUri = _settings.PicBaseUrl;
             var azureStorageEnabled = _settings.AzureStorageEnabled;
 
-            foreach (var item in items)
-            {
-                item.FillProductUrl(baseUri, azureStorageEnabled: azureStorageEnabled);
-            }
+            items.ForEach(item => { item.FillProductUrl(baseUri, azureStorageEnabled: azureStorageEnabled); });
```
Marusyk, line 307: What is the point of the change?

src/Services/Catalog/Catalog.API/Extensions/CatalogItemExtensions.cs
```diff
@@ -6,9 +6,12 @@ public static void FillProductUrl(this CatalogItem item, string picBaseUrl, bool
         {
             if (item != null)
             {
-                item.PictureUri = azureStorageEnabled
-                   ? picBaseUrl + item.PictureFileName
-                   : picBaseUrl.Replace("[0]", item.Id.ToString());
+                // https://www.linkedin.com/feed/update/urn:li:activity:6841055191211491328/
```
Marusyk, line 9: Do we really need this link to the Linkedin post here? It didn't answer the question what is better to use

src/Services/Identity/Identity.API/Services/ProfileService.cs
```diff
@@ -21,6 +21,8 @@ public ProfileService(UserManager<ApplicationUser> userManager)
             _userManager = userManager;
         }
 
+        //New Method ArgumentNullException.ThrowIfNull (obj) Added to .NET 6
```
Marusyk, line 24: I'm sure this source code is not a guide for C# new features, it is more about architecture.
Why do we need to include links to posts? Why not official documentation then?

Marusyk (changes requested):

PR #1759, opened 2021-09-20
Updates set-value package to 4.0.1

sughosneo:
This PR updates `set-value` package version to `4.0.1`


sughosneo:
Tested locally and LGTM!

PR #1809, opened 2021-12-20
fix constructor injection for _validators

A-Soltani:
Array can't be used for constructor injection in ValidatorBehavior class and it makes a runtime error.
IEnumerable is the right type for fixing this issue.


erjain:
Hi @IEvangelist, can you confirm the changes proposed in this PR.

IEvangelist (approved):

Issue #1845, opened 2022-02-04
Eshop page shows Error requesting catalog products, please try later on 


jiheechoi01:
I'm having issues with my eshopOnContainers page. Please see the errors below: 
All the pods are successfully running, however when I go to the http://eshop.3d826a4386af433db850.eastus.aksapp.io/
**1)**  I see the "Error requesting catalog products" for both public and login 
![catalog error](https://user-images.githubusercontent.com/82827230/152607058-5e2c3a38-742e-4dad-8a8e-6148f6af69ea.jpg)

**2)** http://eshop.3d826a4386af433db850.eastus.aksapp.io/catalog returns unhandled exception
![catalog erro 2](https://user-images.githubusercontent.com/82827230/152607128-bb316862-63a1-4902-a3d3-29c33366c942.jpg)

**3)** When I go to webmvc page http://eshop.3d826a4386af433db850.eastus.aksapp.io/webmvc, everything loads up successfully and I can login, add items to basket, and complete an order. However all the images of each item doesn't load up. 
![image](https://user-images.githubusercontent.com/82827230/152607260-1ca27105-ae8c-493f-9a5d-19bf1fdc4c26.png)

Would someone be able to assist in these issues?

erjain:
Hi @jiheechoi01 ,  thank you for reaching out and reporting the issue. We have already identified the issue last week and working on it to resolve it asap. We will update you on this thread once after resolving this bug.

Dintwe:
of course running everything on my dev machine, I have switched the gateway urls to use my local IP, maybe something to consider to see if the requests are going through.

<img width="998" alt="image" src="https://user-images.githubusercontent.com/56117627/153555837-cf29a5f1-3c0b-46de-acb8-92a7f06c0587.png">

it pulls the catalog, but not on first attempt, I had to F5 more than once though...strange.

<img width="1253" alt="image" src="https://user-images.githubusercontent.com/56117627/153552633-e3f7f511-612e-4004-9b1e-cca9959a53f8.png">

and mvc web loads as well

<img width="1217" alt="image" src="https://user-images.githubusercontent.com/56117627/153555569-580c545c-afb2-4822-8cc2-be2ad6918d7c.png">

Dintwe:
...and also for mobile had to make some edits here:

<img width="578" alt="image" src="https://user-images.githubusercontent.com/56117627/153559287-8d910216-23dc-49ed-90c3-e24da989d55b.png">



[...]

PR #1857, opened 2022-02-15
Updated envoy version for Webshopping API Gateway

nishanil:
Fixes the route issue mentioned in the https://github.com/dotnet-architecture/eShopOnContainers/issues/1845


erjain:
Fixes worked for me also on local k8s and AKS both.

PR #1875, opened 2022-03-14
Update appsettings.json

marranica:
Fixed typo in Vault ClientId configuration value from "your-clien-id" to "your-client-id"


nishanil:
@marranica Good catch! Thanks for the fix.

PR #1894, opened 2022-04-10
Removing duplicated AutoAuthorizeMiddleware

countneuroman:


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet-architecture/eShopOnContainers?pullRequest=1894) <br/>All CLA requirements met.

erjain:
Hi @countneuroman, It's not duplication, technically it's microservice's nature, ideally, these repositories will remain separate, and we don't want to have cross-reference into these projects, I don't think we should merge and keep the single reference of AutoAuthorizeMiddleware.

countneuroman:
Thanks for the feedback!

PR #1900, opened 2022-04-20
Remove autofac from the solution

adriver-kwiboo:
Removed Autofac from the solution and replaced with IServiceScopeFactory.
There is a new extension method "AddEventHandlers" that when provided with an assembly will add all implementations of the IIntegrationEventHandler


erjain:
Hi @adriver-kwiboo, Thanks for the contribution. I am keen to know why you are removing Autofac from the solution and replacing it with IServiceScopeFactory, and how it will optimize the source code?

adriver-kwiboo:
Hi @erjain, it was my thinking that as a sample reference application to introduce people to dotnet and architecting complex solutions; reducing the number of external dependencies and using the inbuilt framework it would improve adoption and reduce the barrier for learning.
i.e. if someone that hasn't used Autofac before, but is aware of dotnet and the inbuilt DI, the patterns are going to be more familiar and they can focus on the actually implementation, rather than having to research how Autofac works and what it's role is within the solution.

I'm not sure if there is any optimization in terms of performance, maybe that is something that could be benchmarked.

erjain:
Hi @adriver-kwiboo, you can fork a separate branch for now and keep the changes for reference, right now we do not have much bandwidth to go for this update and review, in case there is any ask from the community regarding this, we will surely try to refer you. Thanks again for your effort. :)

PR #1892, opened 2022-04-08
Order.API project MediatR error solved.

onurkanbakirci:
In the order service's Order.API project, there is a 
`CS0314: The type 'TRequest' cannot be used as type parameter 'TRequest' in the generic type or method 'IPipelineBehavior<TRequest, TResponse>'. There is no boxing conversion or type parameter conversion from 'TRequest' to 'MediatR.IRequest<TResponse>'.`
error. After adding 
`where TRequest : IRequest <TResponse>`
code block to **TransactionBehaviour**, **ValidatorBehavior**, **LoggingBehavior** classes generic type constraints, this error solved.

PR #1911, opened 2022-05-06
MediatR error solved.

onurkanbakirci:
I accidentally closed the topic #1892. This pull request continuation of topic #1892. After reviewing error, i figure out error is about version incompatibility of MediatR. I currently use MediatR 10.0.1 and MediatR.Extensions.Microsoft.DependencyInjection 10.0.1 in my project. After downgrade my MediatR version from 10.0.1 to 9.0.0 (currently used in dotnet-architecture/eShopOnContainers), error was solved. So, if eShopOnContainers will use MediatR 10.0.1, where `TRequest : IRequest <TResponse>` code snippet have to added all of behavior classes inside of src/Services/Ordering/Ordering.API/Application/Behaviors/ path.


erjain (approved):
Thanks for updating the version @onurkanbakirci, I have tested this PR.

PR #1956, opened 2022-08-19
minor refactoring of Webhooks.API

Ich1goSan:
1) fixes in (typo, ctor, unused fields, indents)
2) fixed comparing of hosts in GrantUrlTesterService

### DIFF
--- a/src/Services/Webhooks/Webhooks.API/Controllers/WebhooksController.cs
+++ b/src/Services/Webhooks/Webhooks.API/Controllers/WebhooksController.cs
@@ -69,8 +69,10 @@ public class WebhooksController : ControllerBase
             await _dbContext.SaveChangesAsync();
             return CreatedAtAction("GetByUserAndId", new { id = subscription.Id }, subscription);
         }
-
-        return StatusCode(418, "Grant url can't be validated");
+        else
+        {
+            return StatusCode(418, "Grant url can't be validated");        
+        }
     }
 
     [Authorize]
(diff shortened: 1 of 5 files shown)


ravividap (approved):
Looks good.

erjain:
Hi @Ich1goSan, Thanks for raising the PR, this is working fine, but there is one doubt.

src/Services/Webhooks/Webhooks.API/Controllers/WebhooksController.cs
```diff
@@ -69,10 +69,8 @@ public async Task<IActionResult> SubscribeWebhook(WebhookSubscriptionRequest req
             await _dbContext.SaveChangesAsync();
             return CreatedAtAction("GetByUserAndId", new { id = subscription.Id }, subscription);
         }
-        else
-        {
-            return StatusCode(418, "Grant url can't be validated");
-        }
+
+        return StatusCode(418, "Grant url can't be validated");
```
erjain: why you have removed this statement from the else block?

Ich1goSan: Hi @erjain. I removed that statement because there is an If block above with a return statement. If it's not obvious, of course, I can return it as it was. The same logic with statements like this applied for example here https://github.com/dotnet-architecture/eShopOnContainers/blob/a339e213ba8aebfaeb88e6c1ad0bb7fc0e80725a/src/Services/Webhooks/Webhooks.API/Controllers/WebhooksController.cs#L80-L93

erjain: Hi @Ich1goSan, thanks for the clarification, but I believe this condition will be required to return the proper StatusCode when grantOk is false. Let's revert this and proceed with other changes.

Ich1goSan: @erjain Okay, reverted that one :)

erjain (approved):

PR #1960, opened 2022-08-27
Remove unused ProfileService

### LINKED ISSUE #1951: ProfileService is not used
I don't see `ProfileService`/`IProfileService` used anywhere.

I was looking for an example of how to use permissions in DDD and I noticed that these types aren't used anywhere.

MisinformedDNA:
Fixes #1951


ravividap (approved):
Looks good!

erjain:
Hi @MisinformedDNA, I am getting errors after running with the changes. Can you test the complete application workflow?

MisinformedDNA:
I opened the code in Codespaces, but running `docker-compose build` followed by `docker-compose up` results in DB errors for multiple/all(?) databases like:

> Microsoft.Data.SqlClient.SqlException (0x80131904): Cannot open database "Microsoft.eShopOnContainers.Service.IdentityDb" requested by the login. The login failed. Login failed for user 'sa'.

erjain:
Hi @MisinformedDNA, Sometimes an error during startup can disturb the application deployment and entries in the databases. Could you please clean up old docker images and build the docker images again? - To run the app 
- Go to the `/src/` directory

```text
docker-compose build
```

```text
docker-compose up sqldata
```
Once it's ready to accept the new connection, then you can bring up the other container by using the following command from a separate terminal window. 

```text
docker-compose up
```

Hope it works for the original code.

Ich1goSan:
I think we can't remove ProfileService because it's called when IdentityServer needs to return claims about a user to client applications. IdentityServer calls it internally (the implementation of IProfileService) so we can't see that this class is used.

In the code below these claims are retrieved to UserData field
https://github.com/dotnet-architecture/eShopOnContainers/blob/8438a6a652ce7db4e271e7cd47071fc4d58fdb8c/src/Web/WebSPA/Client/src/modules/shared/services/security.service.ts#L54-L76

Then used below
https://github.com/dotnet-architecture/eShopOnContainers/blob/8438a6a652ce7db4e271e7cd47071fc4d58fdb8c/src/Web/WebSPA/Client/src/modules/orders/orders.service.ts#L51-L91

erjain:
Hi @Ich1goSan, you are right. Thanks for bringing this to the attention here. I am closing this PR and related issue.

PR #2078, opened 2023-03-29
fix (build): Fixed local dns configuration

cgasis:
updated library manager package version
updated local kubernetes dns
added seq


cgasis:
Disregard

PR #2085, opened 2023-04-12
Commit to migrate payment api to WebApplicationBuilder

### LINKED ISSUE #2093: Payment Api: Migrate to WebApplication Builder

erjain:
migrate payment api to WebApplicationBuilder
remove startup.cs
add editorconfig


src/Services/Payment/Payment.API/Payment.API.csproj
```diff
@@ -35,4 +35,12 @@
     <ProjectReference Include="..\..\..\BuildingBlocks\EventBus\IntegrationEventLogEF\IntegrationEventLogEF.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
DreamRunnerMoshi: This change is not needed.

src/Services/Payment/Payment.API/Program.cs
```diff
@@ -1,14 +1,98 @@
-﻿var configuration = GetConfiguration();
+﻿var appName = "Payment.API";
+var builder = WebApplication.CreateBuilder(new WebApplicationOptions
+{
+    Args = args,
+    ApplicationName = typeof(Program).Assembly.FullName,
+    ContentRootPath = Directory.GetCurrentDirectory()
```
ReubenBond, line 6: Is all of this necessary? You should be able to pass `args` into `CreateBuilder`, the rest should not be needed.

erjain: ContentRoot was also specified in BuildWebHost method of the original Program.cs file, so I added it here also, so as to avoid any breaking changes.

jamesmontemagno (approved):

Issue #2070, opened 2023-03-22
EPIC: eShop Modernization/Refresh


adityamandaleeka:
Creating a top-level issue for us to track the ongoing work to modernize/refresh eShop.

- [x] https://github.com/dotnet-architecture/eShopOnContainers/pull/2064
- [x] https://github.com/dotnet-architecture/eShopOnContainers/issues/2069
- [x] https://github.com/dotnet-architecture/eShopOnContainers/issues/2076
- [x] https://github.com/dotnet-architecture/eShopOnContainers/issues/2077
- [x] https://github.com/dotnet-architecture/eShopOnContainers/issues/2093
- [x] https://github.com/dotnet-architecture/eShopOnContainers/issues/2094
- [x] https://github.com/dotnet-architecture/eShopOnContainers/issues/2095
- [x] Remove superfluous `UseDeveloperExceptionPage` calls (it's the default)
- [x] Fix DataProtection issues by configuring a data protection provider
- [x] Replace Serilog with Microsoft.Extensions.Logging.Console
- [x] Use `ILogger<T>` instead of injecting `ILoggerFactory`
- [x] Replace Autofac with simple Microsoft.Extensions.DependencyInjection usage
- [ ] #2108
- [ ] Migrate from ad-hoc `IConfiguration` usage to strongly typed Options
- [ ] Deduplicate use of gRPC and HTTP by using gRPC only for internal-facing microservices
- [ ] Migrate to .NET container build (when we can support the *.NET container build + docker compose* scenario) and delete Dockerfiles

adityamandaleeka:
cc: @erjain @ReubenBond

ReubenBond:
<details>
<summary>EDIT: Merged into top comment</summary>
* [ ] Migrate from ad-hoc `IConfiguration` usage to strongly typed Options
* [ ] Replace Autofac with simple Microsoft.Extensions.DependencyInjection usage
</details>

ardalis:
Why would you replace Autofac with something that has fewer features?

Also regarding IOptions be sure to read this first: https://www.dabrowski.space/posts/asp.net-options-why-you-should-not-use-it/

ReubenBond:
<details>
<summary>EDIT: Merged into top comment</summary>
[...]

PR #2107, opened 2023-05-09
Modernization

ReubenBond:
This is a joint effort between @davidfowl & myself to modernize and clean up eShop.
There is more to do, but we wanted to merge what we have so far before we continue.

@erjain, we have tested this locally in VS (launch & manually test + automated tests) and via docker compose, but we haven't tested beyond that (Kubernetes, Azure). I don't expect you to review this because the PR is far too large, but if you could run some of your own tests, that would be greatly appreciated.

Addresses #2070


erjain:
This is great work, you guys nailed it. I'll review and test it.

ReubenBond:
@erjain merge when ready

erjain (approved):
Looks good to me.

markheath:
I think this update means that logs are no longer appearing in Seq. Is this intentional? There still is a Seq container in `docker-compose.yml` but its not picking anything up

davidfowl:
Yes it’s intentional. We’re also not done with the clean up. This was just the first big set of changes

PR #2123, opened 2023-06-13
Cache JsonSerializerOptions

eerhardt:
According to https://github.com/dotnet/runtime/issues/65396, using a new JsonSerializerOptions every time JsonSerializer is invoked is a suboptimal pattern.

Fix this by caching JsonSerializerOptions instances.

cc @davidfowl @ReubenBond @eiriktsarpalis

### DIFF
--- a/src/Services/Services.Common/JsonHelper.cs
+++ b/src/Services/Services.Common/JsonHelper.cs
@@ -1,12 +1,11 @@
 ﻿using System.Text.Json;
 
-namespace Services.Common
+namespace Services.Common;
+
+public static class JsonHelper
 {
-    public static class JsonHelper
+    public static readonly JsonSerializerOptions CaseInsensitiveOptions = new()
     {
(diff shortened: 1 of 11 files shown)


src/Services/Services.Common/JsonHelper.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Text.Json;
+
+namespace Services.Common;
+
+public static class JsonHelper
```
davidfowl: nit:
```suggestion
public static class JsonDefaults
```

eiriktsarpalis (approved):

src/Services/Services.Common/JsonHelper.cs
eerhardt: I was thinking `JsonHelper` would leave it open to more things than "defaults" in the future. But we can rename it if/when that happens.

ReubenBond (approved):

davidfowl (approved):

PR #2124, opened 2023-06-15
Clean up usings and controller attributes

davidfowl:
- Removed redundant attributes when the controllers already specifies what the result type is.
- Use StatusCodes instead of HttpStatusCode.
- Clean up namespaces and use type aliases to disambiguate the many DTOs defined.


src/Services/Webhooks/Webhooks.API/Controllers/WebhooksController.cs
```diff
@@ -71,14 +71,14 @@ public async Task<IActionResult> SubscribeWebhook(WebhookSubscriptionRequest req
         }
         else
         {
-            return StatusCode(418, "Grant url can't be validated");
+            return StatusCode(StatusCodes.Status418ImATeapot, "Grant url can't be validated");
```
ReubenBond: I assume the status code used is for fun, but the error message reads wrong.

```suggestion
            return StatusCode(StatusCodes.Status418ImATeapot, "Grant URL invalid");
```

ReubenBond (approved):

PR #2126, opened 2023-06-15
Fixed the spa project

### LINKED ISSUE #2121: CORS-related errors when loading SPA version of web app
I cloned the dev branch and followed the instructions to build and run the project using Docker Compose. After running `docker-compose build` and `docker-compose up`, health checks indicate that all services are healthy.

webstatus

The MVC version of the web app is functioning correctly, but I encountered CORS-related errors when trying to load the SPA version. This issue is likely due to Cross-Origin Resource Sharing (CORS) restrictions.

webmvc

webspa

I would appreciate guidance on how to configure the necessary CORS settings to resolve this issue. Please provide instructions or suggestions on how to modify the project's configuration files or any other relevant steps to enable proper CORS handling for the SPA version of the web app.

Thank you for your assistance.

Please feel free to adjust the content as needed.

davidfowl:
- Clean up the SPA project (removed dead code)
- Fixed URLs in override file

Fixes https://github.com/dotnet-architecture/eShopOnContainers/issues/2121


src/docker-compose.override.yml
```diff
@@ -276,13 +276,13 @@ services:
     environment:
       - ASPNETCORE_ENVIRONMENT=Production
       - ASPNETCORE_URLS=http://0.0.0.0:80
-      - Identity__Url=http://${ESHOP_EXTERNAL_DNS_NAME_OR_IP}:5105
-      - PurchaseUrl=http://${ESHOP_EXTERNAL_DNS_NAME_OR_IP}:5202
+      - IdentityUrl=http://${ESHOP_EXTERNAL_DNS_NAME_OR_IP}:5105
+      - PurchaseUrl=http://${ESHOP_EXTERNAL_DNS_NAME_OR_IP}:5121
```
davidfowl, line 280: This is the real cors fix (going through the bff instead of envoy). I'll drop envoy from a future commit.

src/Web/WebSPA/Program.cs
```diff
@@ -81,16 +24,12 @@
...
-});
-app.MapHealthChecks("/hc", new HealthCheckOptions()
+
+#pragma warning disable ASP0014 // Suggest using top level route registrations
+app.UseEndpoints(routes =>
 {
-    Predicate = _ => true,
-    ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse
+    // TODO: Change this route
+    routes.MapGet("/home/configuration", (IOptions<AppSettings> options) => options.Value);
```
ReubenBond, line 32: This changed from `IOptionsSnapshot<T>` to `IOptions<T>`. In practice I don't think they'll ever change at runtime, but if we are setting best practice, maybe we should allow for that.
```suggestion
    routes.MapGet("/home/configuration", (IOptionsSnapshot<AppSettings> options) => options.Value);
```

ReubenBond (approved):

src/Web/WebSPA/Program.cs
ReubenBond: `IOptionsSnapshot<T>` is inefficient in the common case (where options do not change). We could use `IOptionsMonitor<T>` instead, but these options will not change in practice, and the code clutter is not worthwhile for the approximately zero benefit it would provide.

PR #2128, opened 2023-06-16
Rollback API deleted by mistake on Webhook Client

RobertoBorges:


RobertoBorges (commented):
@davidfowl I saw a PR where you removed this API, you were cleaning stuff, I think you removed that by mistake and the Webhook Client was not working anymore... I'm just rolling back that

davidfowl:
Thanks @RobertoBorges, we should clean this up if it is indeed needed. I will leave feedback on the PR.

PS: Also, there are not tests for this part of the project 😞. Where is this endpoint used?

RobertoBorges:
> Thanks @RobertoBorges, we should clean this up if it is indeed needed. I will leave feedback on the PR.
> 
> PS: Also, there are not tests for this part of the project 😞. Where is this endpoint used?

This end point is used on the Webhook process where the subscription request a confirmation at this end point... Without it the Subscription verification by the WebHook API don't work when registering the WebHook Client.

 this was the commit that removed the endpoint:

https://github.com/dotnet-architecture/eShopOnContainers/commit/df4eb4c124f9c656ff63ba3eef93e4704961add2#diff-8200407cc237c6fe9940cb6c2334070720f88bcd54b739d001bd8e51404487d5

davidfowl:
OK finally getting around to understanding this better and I will put it back

PR #2143, opened 2023-08-11
BasketItem - replace list by yield return the validation error result

AhmedKhames:


AhmedKhames:
@dotnet-policy-service agree

Issue #2071, opened 2023-03-23
Identiy fails when using externalDNS with tls on Kubernetes


jmasengesho:
Looks like the identity fails when the traffic is coming from htts instead of http. I am running the app on Kubernetes with externalDNS configured and TLS. I am not using LetsCrypt. I brought my own certs.  See screen below the logs from identity pod:

<img width="788" alt="image" src="https://user-images.githubusercontent.com/1504610/227254151-841c024f-24e5-4f47-8a3e-111af12b2000.png">

PR #2161, opened 2023-09-21
update k8s charts; Use ForwardedHeaders;

maddadder:
- Update the k8s charts to reflect the build.  
- Address the issue reported in #2071
- Add an option to use istio in the k8s charts
- fix signalr so that orders show up after placing


maddadder:
@dotnet-policy-service agree

